PR #1, opened 2015-08-13
tests: Skip TestSchedAssert if sheye couldn't be imported

JaviMerino:


sinkap:
Hi Javi, 

I will be changing the sheye namespace to bart.sched, will make these changes and then we can test on bart.sched or just bart.

JaviMerino:
Updated to use bart.sched

PR #2, opened 2015-08-14
Adding Missing Pivots, Pivoting for grammar

sinkap:


sinkap:
The new patch that I added to the tests should fix the failures. The pivoting added requires an extra reference that was breaking the test cases.

sinkap:
I am merging this as the tests are passing now.

PR #4, opened 2015-08-21
plotter: Add script to publish interactive plots

sinkap:
This is a script to publish a notebook containing Ipython graphs
The static data is published as an anonymous gist. GitHub does not
allow easy deletions of anonymous gists.


sinkap:
Will create a new one..this needs some fixes as well..

PR #5, opened 2015-08-25
plotter: Fix profile dir calculation

sinkap:
The older method does not work for IPython 4.0+. Apparently,
there is a better way to get the directory which works for both


sinkap:
Needs more fixes too

PR #7, opened 2015-08-26
Fix traces (re)generation

derkling:
Here is a refracting of the trace generation function which should:
1. make the code a little bit more clear using a common naming schema for trace locals
2. introduce support for text traces re-generation once the corresponding binary trace has been updated

The last feature is especially useful when we run multiple time a test which keep updating the same binary trace file.

The new proposal should diverge from original code just regarding "raw" traces generation, more details in the commit message for the second patch.

### DIFF
--- a/trappy/run.py
+++ b/trappy/run.py
@@ -103,35 +103,35 @@ classes are parsed.
         """Process the path and return the path to the trace text file"""
 
         if os.path.isfile(basepath):
-            trace_txt = os.path.splitext(basepath)[0]
-            raw_trace_path = trace_txt + ".raw.txt"
-            trace_path = trace_txt + ".txt"
+            trace_name = os.path.splitext(basepath)[0]
+            trace_raw = trace_name + ".raw.txt"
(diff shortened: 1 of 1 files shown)


derkling:
I'm going to update and simplify a little bit the second patch, thus dropping this pull request.

derkling:
Well, the "simplification" attempt was not successful, thus opening again the review request in the current status.

derkling:
I've updated the branch to fix the generation of RAW traces when missing while a TXT trace is already available.
Checked with nosetests, all passing... but we are actually missing a test for traces "timestamp" dependencies. I've checked with a notepad and it seems working fine for both files and dirs basenames.

JaviMerino:
We have a test for the timestamp dependency, I added it in a346dfb0593d062df35a171f18ab3e3ed3508831.  I'll rebase your branch on top of the current master to check.

JaviMerino:
Ideally the test for the timestamp should come with the patch...

derkling:
Replaced by v2 which adds a test case as well.

PR #10, opened 2015-09-02
doc: Add IPlotter example

JaviMerino:


sinkap:
Would be nice to run this notebook through publish_interactive_plots.py so that the plots are globally accessible.

JaviMerino:
Sure, can you do it?  We will have to export it as HTML if we want the world to be able to see it from github itself.

sinkap:
I have created a pull request #16 for this.

JaviMerino:
Ok, so this can be closed as we won't merge it, no?

sinkap:
Yes, this can be closed.

PR #20, opened 2015-09-09
Fix traces generation/v2

JaviMerino:


JaviMerino:
This is a follow up of #7.  There's a test to make sure we update the trace.txt if the trace.dat is newer, use `getmtime()` instead of `getctime()` and the simplification I said in https://github.com/derkling/trappy/commit/42363c692741ca6c8b95fcbd4a45b80d8afd7e5b#commitcomment-13023611

@derkling are you ok with this?

derkling:
That seems good to me, thanks for the further simplification.

JaviMerino:
Thanks for doing this, it had been in my todo list for a while :heart:

PR #21, opened 2015-09-09
TRAPpy Documentation Project with Sphinx

sinkap:
I have created a single commit for each module. It makes the review easier and would also allow us to retain history if files are moved around. 

Feel free to add on with suggestions and also any file that I may have missed and let the review begin :+1: 
Once this is merged, I will create an HTML build and add a link to the TRAPpy homepage.


JaviMerino:
The last bits of it were merge in #31

PR #23, opened 2015-09-11
Scatter plots for ILinePlot and LinePlot

sinkap:
I would actually like to add Dietmar as a watcher. He is not on github yet..


JaviMerino:
:heart: lgtm, I'm merging it.  If he wants an improvement, we can add it afterwards.

PR #27, opened 2015-09-18
Documentation v2

sinkap:
Making the required changes.


JaviMerino:
The only patch that was left of this was merged in #31

PR #30, opened 2015-09-22
Iplot install fix

sinkap:
The iplot_install has been moved to the view method to do a lazy installation. This should allow us to import these modules from outside an IPython notebook


JaviMerino:
I haven't tested it. I'm assuming you have and stuff still works.  That's something that I don't think we're testing in the testsuite

sinkap:
Yes, this works for me. We can merge it.

JaviMerino:
Merged in #34

Issue #22, opened 2015-09-11
test_copyright should ignore files in .gitignore


sinkap:
test_copyright should consider files in .gitignore before walking the directory. This is needed for the _build directory in the new API docs which is a part of .gitignore

JaviMerino:
That's why we were using `git ls-files` :wink:. I guess we'll have to replicate that part of the functionality.

sinkap:
Ahh!! Then my patch needs to be augmented. Thanks! Will add this bit as
well.

On Fri, Sep 11, 2015 at 2:20 PM, Javi Merino <notifications@github.com>
wrote:

> That's why we were using git ls-files [image: :wink:]. I guess we'll have
> to replicate that part of the functionality.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ARM-software/trappy/issues/22#issuecomment-139544112>.
>



-- 
Regards,
Kapileshwar Singh(KP)

sinkap:
Fixed by #33

PR #31, opened 2015-09-22
doc: Add Sphinx documentation

sinkap:
Have updated the reference targets

#30 should fix the import error we are getting when we build the documentation (This error only occurs on if IPython version v3 is used, but we needed a fix nevertheless) :hamburger:


JaviMerino:
Ok, I'm happy with it, but I'd like to merge first a fix for #22 (I should send a pull request for that later today)

JaviMerino:
FTR, the pull request for a fix for #22 is #33.

PR #32, opened 2015-09-22
Iplot install fix v2

sinkap:
Made the required changes from #30. This works on IPy v4 but we need to test on IPy v3. I can confirm this later today and merge it.


sinkap:
I have assigned @JaviMerino I can do the test on IPy v3 if that is fine?

JaviMerino:
Check your testing, if you removed the scripts, you should only have `EventPlot` in `~/.ipython/profile_default/static/plotter_scripts` with this version.

JaviMerino:
Merged in #34

PR #34, opened 2015-09-23
Iplot fix v3

sinkap:


JaviMerino:
Tested with ipython 2 and 3

PR #36, opened 2015-09-23
API Reference on gh-pages for TRAPpy

sinkap:


JaviMerino:
It's a bit of a hack, but the best we can do right now.  Hopefully we will move it to pythonhosted soon and we can get rid of the static build.

sinkap:
That would be ideal :+1:

PR #39, opened 2015-10-09
User specified lanes for EventPlot

sinkap:


sinkap:
I need to change my ipython version to 3.0.0 or have a check that prevents me from checking in v4 notebooks.

JaviMerino:
Maybe you can add the check to =hook/pre-commit=

JaviMerino:
This was merged as #43

Issue #45, opened 2015-10-27
trappy.plotter.plot_trace should accept Run objects


sinkap:
It currently accepts only paths which is painful

JaviMerino:
Merged as #60

JaviMerino:
Merged as #60

PR #51, opened 2015-11-02
plotter: Accept run objects in plot_trace

sinkap:
To avoid the recreation of the run object for large trace
files. Enhancement #45


JaviMerino:
Merged as #60

PR #63, opened 2015-11-06
Fix #48 trappy.Run() should allow to parse just a subtrace

### LINKED ISSUE #48: [FEATURE] trappy.Run() should allow to parse just a subtrace
It could be useful in some situations to parse only a portion of the trace.
For example, to consider only the events in between two specified markers or the events from a certain (relative) initial timestamp and for a certain duration.

This for example cold speed/up some post processing analysis which starts from a complete trace but then they need to focus on specific intervals.

JaviMerino:


JaviMerino:
I've updated the branch instead of pushing a new one because it looks like github preserves the comments of the previous patches when you push new rebased patches.

JaviMerino:
new version with relative and absolute windows

sinkap:
Excellent Stuff! :+1: :tada: Works like a charm!

PR #65, opened 2015-11-11
Added 3 new plotters

ep1cman:
XYPlot:
 A simple wrapper around matplotlib thats allows compositing more complex figures

DualPlotNorm:
 Plots two side by side graphs of the same data. One with error bars the other normalised to a percentage

PowerVsPerf:
 Generates a Power Vs. Performance graph from a WA directory


sinkap:
I have uploaded a slightly modified notebook as a gist at

https://gist.github.com/sinkap/9b043c362c4f25f49818

which shows how most of the functionality in XY charts can be used in LinePlots. We need to think about error bars in LinePlots.

derkling:
@sinkap what is the status of these plotters? I had a look at the example notebook you posted and it seems really promising! ;-)

sinkap:
@derkling I am not sure where we are with this. We might want to add XY plotting to the existing LinePlotter and also add error bars to it.

JaviMerino:
This is now in [PyYams](https://github.com/ep1cman/PyYams).

PR #67, opened 2015-11-11
plotter: add support for figure title

derkling:
The LinePlotter can generate one or many plots, depending on pivoting
and filters configuration. All these plots usually provides a common view
on a specific set of metrics and/or experiment.
An overall figure title could be useful to better represent what the
set of generated plots refers to.

This patch adds the required support to defined a figure title which can
be defined using an optional new 'title' paramter. For example:

trappy.LinePlot(run, pivot="cpu", title="This is a title").show()


sinkap:
Cheerio! :tada: :wine_glass:

PR #71, opened 2015-11-16
run: sched: rename "state" columns for cpu_frequency and cpu_capacity

derkling:
CPU Frequency and Capacity trace events report the actual "frequency" and
"capacity" using an anonymous "state" column. Plots generated using Plotter
use that columns name in the legend, thus making a bit confusing the plot,
even worst when we want to plot both frequency and capacity.

This patch assigns some more meaningful name to the "state" report by
these two trace events.


sinkap:
This is still valid right?

Issue #69, opened 2015-11-13
[FEATURE] Extend Run object constructor to get a list of custom events


derkling:
The FTrace module provided by devlib allows the user to specify a simple list of events to be collected via FTrace. These events and only them are usually the interesting events for the following analysis.

For that reason, the support in Run() for a new "custom" scope could be useful to better match the devlib API. A new parameter "events" should allows to dynamically register at Run() object construction time only the list of events specified by the user.

This is expected to optimize memory and parsing time, while also solving some handshakes due to mixed usage of pre-defined and dynamically registered events.

derkling:
I've noticed that some events (e.g.  sched_switch) require a raw trace.
Scheduler switching events are always parsed by trappy to support the trace plotting API, thus it make sense to keep it pre-registered.

However, in general, if we want to introduce a new support for "list based" events definition we should also think about a convenient way to defined which events are to be parsed from the raw trace and which can be parsed from the "formatted" trace.

JaviMerino:
We could register the existing `trappy` class for things like `sched_switch` instead of doing a `register_dynamic`.  That should take care of the raw traces that `trappy` already knows about.  For those that `trappy` doesn't know about you'll have to register them yourself before as usual.

derkling:
The proposed implementation matches my expectations. Thanks. Looking forward to see it merged.

Issue #68, opened 2015-11-13
Run(scope="custom") shall not register any thermal or scheduler classes 


JaviMerino:
@derkling suggests adding a "custom" (or "dynamic") scope to the run object in which only dynamic classes are parsed.

JaviMerino:
This was fixed in #78

PR #78, opened 2015-11-26
Run events

JaviMerino:
Implement #68 and #69

### DIFF
--- a/doc/Run class.ipynb
+++ b/doc/Run class.ipynb
@@ -1,7 +1,7 @@
 {
  "metadata": {
   "name": "",
-  "signature": "sha256:994dc0a28da6fa742c7f300846942769885060c72f99a4a2550db234d3f499a0"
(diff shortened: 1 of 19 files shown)


JaviMerino:
@derkling If I understand it correctly this is what you wanted.  Can you test it to confirm that it fits the bill?

derkling:
We are almost there, the new API is way more simple to use.
However, we still have some issues due to collisions with pre-registered events.

For example, if I need to trace the "cpu_capacity" event using the new API,  we end up on re-using the pre-defined sched::SchedCpuCapacity class, which is registered as "sched_cpu_capacity".
This misalignment is confusing for the user and thus must be avoided when the new API is merged.

I have a patch which renames all the pre-registered sched related events to match the eventually user defined name. But it's so simple that you can probably just do it on top of your series.
[rename_sched_events_patch.txt](https://github.com/ARM-software/trappy/files/45178/rename_sched_events_patch.txt)

PR #85, opened 2015-11-30
run: provide a name for the class if there isn't one already

JaviMerino:
A number of trace classes are called as the trace event they parse.
Unsurprisingly, the unique_word *is* the trace event name.  Instead of
repeating the same thing over and over, let register_class() assign a
name to a class it's registering based on the unique_word if it doesn't
have a name.

This needs to be a class variable that's set before the class is
instantiated (Run.register_class() needs it).  So you can't set it in
Base's constructor.  You can't set it in DynamicTypeFactory either
because not every Base class uses that.  This looks like a good
compromise and remove some redundant code.

Change-Id: I6389c9ac072f6571b4faf8dacdee4cd77775a1f8

### DIFF
--- a/trappy/run.py
+++ b/trappy/run.py
@@ -229,6 +229,10 @@ class Run(object):
             :mod:`trappy.dynamic.register_dynamic` :mod:`trappy.dynamic.register_class`
 
         """
+
+        if not hasattr(cobject, "name"):
+            cobject.name = cobject.unique_word.split(":")[0]
+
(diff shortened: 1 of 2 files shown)


JaviMerino:
I wish we could change thermal classes (especially `ThermalGovernor`, which has a completely inappropriate name).  But I believe they are already hardcoded in some tests and it will be disrupting to change it for no good reason other than "the name makes more sense" and "we reduce the codebase".

PR #86, opened 2015-12-02
stats: Fix Bugs in grammar

sinkap:
* Incorrect scope for assignment of concatenated index
* missing check for attribute presence


JaviMerino:
Merged as #91

PR #90, opened 2015-12-02
Systrace support

derkling:
That's an initial set of patches which adds support for parsing of textual traces embedded in the
systrace generated HTML report.

The second patch in the series is the most controversial... perhaps we could find a better solution.
Let discuss it after lunch ;-)


JaviMerino:
I quite like this change.  We are working on generalizing the `Run` object to become something that can be subclassed.  Then we could have a `Systrace` class that implements what you've done here.

JaviMerino:
Merged as #160

Issue #89, opened 2015-12-02
trappy should have an API to add dataframes to the run object


JaviMerino:
If `run` is an instance of `Run` and `dfr` is a pandas `DataFrame` then something like:

    run.add_event("important_event", dfr, pivot="cpus")

Should let me do `run.important_event.data_frame`.  Any other trappy or bart class or function that expects a Run object with multiple events will be able to operate on it.

JaviMerino:
Kudos if we let `Run` instances be added to each other:

```
In [1]: import trappy

In [2]: run1 = trappy.Run(events=["thermal_temperature"])

In [3]: run2 = trappy.Run(events=["thermal_power_cpu_limit"])

In [4]: full_run = run1 + run2

In [5]: full_run.thermal.data_frame.head()
Out [5]:
               __comm  __cpu  __pid  id   temp  temp_prev thermal_zone
Time                                                                  
0.000000  kworker/4:1      4     51   0  68786      68794  soc_thermal
0.099539  kworker/4:1      4     51   0  68781      68786  soc_thermal
0.199957  kworker/4:1      4     51   0  68782      68781  soc_thermal
0.299112  kworker/4:1      4     51   0  68835      68782  soc_thermal
0.399646  kworker/4:1      4     51   0  68882      68835  soc_thermal

In [6]: full_run.cpu_in_power.data_frame.head()
Out[6]:
               __comm  __cpu  __pid               cpus  dynamic_power  \
Time                                                                    
0.002264  kworker/4:1      4     51  00000000,00000006            250   
0.002279  kworker/4:1      4     51  00000000,00000039            166   
0.100949  kworker/4:1      4     51  00000000,00000006            227   
0.100970  kworker/4:1      4     51  00000000,00000039            178   
0.200914  kworker/4:1      4     51  00000000,00000006            256   

             freq  load0  load1  load2  load3  static_power  
Time                                                         
0.002264  1100000     24     19      0      0           178  
0.002279   850000     36     49     48      7           131  
0.100949  1100000     28     11      0      0           178  
0.100970   850000     32     28     46     44           131  
0.200914  1100000     20     24      0      0           178

JaviMerino:
With #95 and #117 this is now done.  I didn't go for the extra kudos because I don't see the point of adding `FTrace` classes (previously known as `Run`)

PR #95, opened 2015-12-04
run: attach events to an existing run event

JaviMerino:
Add an "add_parsed_event" method to Run  that lets you add a dataframe
to the events in the run.


JaviMerino:
Not a full fix for #89, but at least it's a start

PR #96, opened 2015-12-04
Fix get_duration for normalized time

sinkap:
get_duration should return the same value irrespective of the
normalization of time. Duration is a relative parameter that
measures the length of time from start to the end of the trace


JaviMerino:
:heart:, thanks for this!

PR #98, opened 2015-12-08
performance improvements

JaviMerino:
A couple of performance improvements to parse the trace when only looking for a handful of events or a small window.


JaviMerino:
Merged as #100

PR #101, opened 2015-12-10
Handle Duplicate Indices

sinkap:


JaviMerino:
Other than that, I'm quite happy with this.

PR #104, opened 2015-12-11
stats: Remove Scheduler specific functions

sinkap:
The scheduler specific functions should be a part of BART


JaviMerino:
I'm happy with this :smile:, I'll merge it when we merge ARM-software/bart#43

PR #112, opened 2015-12-17
LinePlotter: don't make titles appear as t,h,i,s

JaviMerino:
In 23724934e98c ("plotter: Pivoted Runs") we broke the titles of
LinePlotter for plots done with column=string.  After that, the titles
for those plots appeared as a comma-separated list of letters:
l,i,k,e,t,h,i,s.

Fix it so that one-column plots display the column as the title and
multi-columns plots display a comma-separated list of columns as title.


sinkap:
PS: The commit message is excellent!

Issue #24, opened 2015-09-11
plotter: Responsive design for interactive plots


sinkap:
We currently have the width and height of the plots statically defined. This makes them look disproportionately large on mobile/non-standard screens. We can use responsive UI designs to overcome this. 

Not really an issue, but a feature request. :chart_with_upwards_trend:

sinkap:
Done! Finally!

PR #113, opened 2015-12-19
Dynamic resizing for EventPlot

sinkap:
Fixing issue #24 for EventPlots


JaviMerino:
I've tried it with the example notebook in `doc/InteractivePlotter.ipynb`.  The plot in `trappy.plotter.plot_trace()` works ok but the two previous ones (the ones generated with `trappy.EventPlot`) don't show properly if you resize the screen.  It works if you rerun the notebook so I'm merging it, but this may be worth looking at in the future.

sinkap:
Thanks for pointing this out Javi, as it turns out I was overriding the resize event listeners for old graphs. I have pushed a fix for this #114

PR #114, opened 2015-12-20
plotter: EventPlot: Namespace the resize event listeners

sinkap:
Since the listener needs to fire for each plot on the page
we need to add a namespace to the registered listener. Not
doing doing this causes the resize listener to be fired for
the last created plot. The name-spacing can be done as follows

.on("resize.one", function_one)
.on("resize.two", function_two)

### DIFF
--- a/trappy/plotter/js/EventPlot.js
+++ b/trappy/plotter/js/EventPlot.js
@@ -438,7 +438,7 @@ var EventPlot = (function () {
             }
 
             d3.select(window)
-                .on("resize", resize)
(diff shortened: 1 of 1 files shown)


sinkap:
It works for me on safari and Chrome. Can you try the following steps:

* Restart the kernel and clear all outputs
* Clear the cache
* Refresh the page
* Rerun all cells

JaviMerino:
Yep, my bad, it works on firefox as well.

PR #116, opened 2015-12-21
Fixes for #89

JaviMerino:
Generalize the Run object.  Rename it to FTrace and extract the more generic bits into a BareTrace class.  This makes trappy more generic and lets you use it easily with traces other than ftrace.  This enhances #89.

PR #117, opened 2015-12-22
Fix 89 v2

JaviMerino:
Addressing comments in #116


sinkap:
Awesome! All set merging this!

PR #121, opened 2015-12-27
ftrace: fix error with only one actor and add corresponding test case

Leo-Yan:
When system has only one actor and call method trappy.summary_plots, it
will report failure "zip argument #1 must support iteration".
    
This failure is cause by parsing allfreqs plots. Python always pass by
value, so when there have only one actor the axis will pass by value and
it's not a array type anymore. So finally zip will report it cannot
support iteration. So in this case convert axis to array type.

Also per Javi's request, Javi shared one test case for it, so also commit test case for it.

### DIFF
--- a/tests/test_ftrace.py
+++ b/tests/test_ftrace.py
@@ -265,6 +265,22 @@ class TestFTrace(BaseTestThermal):
         trace.plot_allfreqs(self.map_label, ax=axis)
         matplotlib.pyplot.close('all')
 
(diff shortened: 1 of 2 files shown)


JaviMerino:
I've squashed both commits into one and merged it as 8da87510391505576e2b19331f236eab06844316.  We prefer to have the test and the fix in the same commit, I think it makes it clearer.

PR #122, opened 2015-12-31
ftrace: fix error when compare runs

Leo-Yan:
When compare multiple run instances, need convert axis objects from 2D
array type to 1D array type; in the code it use zip() to do the
conversion. But when every run instance has only one actor, then it will
directly return back axis object with 1D array type, so zip function
will report failure.

This patch is to check if axis is 2D array, and will convert it to 1D
array.


JaviMerino:
This should be fixed by #125

PR #124, opened 2016-01-05
setup: install EventPlot_help.jpg

JaviMerino:
When you install trappy using "pip install" or "python setup.py
install", trappy.EventPlot fails to run because it can't find
EventPlot_help.jpg.  Install it with the rest of the data files.


sinkap:
Yes this is the problem that @punitagrawal also faced when he pip installed trappy. Good that it is fixed now :)

JaviMerino:
Correct, I should have mentioned him :disappointed:

PR #126, opened 2016-01-06
Legend outside

John-P:
I've tried to add the ability for the legend to be put the the side of the plot area in a notebook which would be useful for large legends. Unfortunately, at the moment the image is getting cropped so that only half the legend is shown.


mdigiorgio:
@JaviMerino should we close this? #147 already establishes a template for plots

JaviMerino:
yep, we can close this now.

PR #127, opened 2016-01-06
Custom colors

John-P:
Modified LinePlot and ColorMap so that if the 'colors' kwarg is set as a list of rgb values e.g. [[0,255,0]] or [(0,1,0)] then a listed colormap will be made and used instead of the default one. At the moment I have added just added a function to ColorMap that will overwrite the defualt cmap if the colors kwarg is set but it could be changed to be implemented so that the ColorMap object is initialised differently for the default ('hvs') or custom rgb colormaps.


JaviMerino:
Addressed the review and merged as 35e4045.

PR #128, opened 2016-01-14
Moved resolving of constraints from LinePlot to an abstract class

John-P:
Added a StaticPlot abstract class between abstract plotter and LinePlot which handles the resolving of constraints. StaticPlot now has a plot_axis method which is called from StaticPlot once for each axis with a list of the series (lines in this case) to be plotted on that axis.


JaviMerino:
Fixed as #135

PR #129, opened 2016-01-14
Bar plot

John-P:
A branch off of my abstract-constraints branch which adds a new plot which can plot vertical bars.

Series may also be plotted as stacked bars by setting the `stacked` kwarg to true.
Spacing between groups may be set by setting the `spacing` kwarg to a float from 0.0 to 1.0.

Data values are plotted against their position index in the list (rather than pandas dataframe index column) and series are plotted together in groups around each index. 

Also added a fix for plotting data frames which is currently broken in trappy origin.


JaviMerino:
Can you rebase this on top of master?  I've merged the fix for plotting of dataframes.

Usually it's just `git rebase master bar-plot` but in this case you may have conflicts and maybe a stray commit at the end with the unneeded bits of the "fixed plotting of dataframes" commit that you may need to remove.  Let me know if you need help.

JaviMerino:
Merged as f4dbd1a60d428b70ea8fc1f05f59343d1db4900c

PR #132, opened 2016-01-15
Fixed plotting of dataframes

John-P:
This fixes plotting of dataframes for trappy origin. I have also added this fix head to the bar-plots branch which also abstracts the resolving of constraints out of LinePlot and implements a BarPlot. However, I have submitted this separately incase the bar-plot branch is too difficult to merge and the fix is needed quickly.

### DIFF
--- a/trappy/plotter/AbstractDataPlotter.py
+++ b/trappy/plotter/AbstractDataPlotter.py
@@ -39,7 +39,7 @@ class AbstractDataPlotter(object):
         self._event_map = {}
         self._attr = attr if attr else {}
         self.traces = traces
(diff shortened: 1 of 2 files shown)


JaviMerino:
I would usually ask you to resend it with the fix but given that it's taking me too long to review the other pull requests I've fixed this and merged it as 96602fc3cafea4e89a7b938e6d1363add1b9613f.  I've added a test case to prevent from breaking it again :angel:

John-P:
I think I fixed it for breaking signals but maybe i only pushed to the bar-plot branch

PR #136, opened 2016-01-28
base: make parsed_data a generator

JaviMerino:
With big traces, the `parsed_data` array becomes huge (up to gigabytes).  `parsed_data` is only used as a parameter for `pandas.DataFrame`.  pandas accepts a generator instead of a list.  Therefore instead of creating a huge array for pandas, pass a generator to pandas and let it process the incoming information as it needs to.

Before:

__init__-before

After:

__init__-after

As far as I can tell, `.0` and `data` are internal pandas variable and thus outside of our control.

### DIFF
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -157,19 +157,7 @@ class Base(object):
         self.cpu_array.append(cpu)
         self.data_array.append(data)
 
-    def create_dataframe(self):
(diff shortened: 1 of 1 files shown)


sinkap:
Looks good to me! This will surely decrease the memory footprint. We need to use generators more often in other areas of the code. Thanks for starting the memory profiling effort!

PR #138, opened 2016-02-03
sched: add classes for wakeup/wakeup_new events

derkling:
Some recent version of trace-cmd prints sched_wakeup and sched_wakeup_new
events using a format string. Thus, when this events are required we must
parse them from the raw trace.

Since these events are already available mainline, this patch pre-register
them dynamically as raw events. This allows to parse these events (when
required) using the proper trace.

The patch adds also a test case by using the original formatted trace as
a source for the raw trace to use.


JaviMerino:
cool, thanks! :grinning:

PR #140, opened 2016-02-12
Plotter Permute example

mike2390:
Example of LinePlot using the permute parameter.

This PR also includes a patch for a cleaner print of the legend, i.e.:
- No legend when there is only one trace and only one column has to be plotted
- No trace name in the legend if there only one trace


derkling:
Tested last two patches and ("move legend below graph" and "make_title show trace name only for single trace..." on a couple of notebook of mine and they are working properly.
Plots are not much more "sexy"... it would be great if we can have the same updates for the ILinePlot class ;-)

JaviMerino:
Merged :smiley:

PR #142, opened 2016-02-23
Constraint fixes

JaviMerino:
A number of simplifications of the Constraint code.  The most important is the first one that fixes a bug for which with some traces we raise a "Unable to handle duplicates" where we shouldn't.

### DIFF
--- a/tests/test_plotter.py
+++ b/tests/test_plotter.py
@@ -67,7 +67,7 @@ class TestPlotter(BaseTestThermal):
                           trace2],
                          [trappy.cpu_power.CpuOutPower],
                          column=["power"],
-                         filters={"cdev_state": [1]})
+                         filters={"cdev_state": [0]})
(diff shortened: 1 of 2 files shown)


JaviMerino:
@deggeman The first patch should fix your problem regarding plots with multiple filters that were failing with "Unable to handle duplicates".  Can you please test it?

deggeman:
Tested on a ipynb which uses lots of filters with multiple entries in trappy.ILinePlot(..., filters = { ... }, ...). No problems encountered :-)
(It was failing before and the only workaround was to lower max_delta in handle_duplicate_index()?)

sinkap:
Only one comment about handle_duplicate_indices. Otherwise, if @deggeman is happy, so am I :smile:

JaviMerino:
Merged as #146

PR #144, opened 2016-02-24
Teach the Parser to apply a window to data accesses

JaviMerino:


sinkap:
Thanks for adding this. Looks like we are starting to leverage the grammar now! Good to see :+1:

JaviMerino:
yep, I'm improving a test in lisa - making it more "barty" :wink:  Next step: filters for `Parser()` similar to the ones in plotter.  Actually, I should create an issue for that.

PR #147, opened 2016-02-26
Uniform template for static and interactive plots

mdigiorgio:
This series uniforms the style of both static and interactive plots. The template is the following:

template_plots

Where:

- The legend is always shown below the graph, centered with respect to the plot. Notice that in case of static plots we just need a single legend.
- **Master title** is optional. It is shown only if the user provides one.
- **TraceName** in the legend is not shown when is already in the title.
- Background for plots is white.
- Static plots legends have light grey background to better distinguish the labels.

For the interactive plots, there's no possibility to tell the `dygraph` class how many labels to put in the legend, but it is possible to have them one per line. Therefore, they are shown in this way to avoid overlapping of labels when plotting many traces.

### DIFF
--- a/trappy/plotter/AttrConf.py
+++ b/trappy/plotter/AttrConf.py
@@ -54,7 +54,7 @@ MPL_STYLE = {
                          '#188487',
                          '#E24A33'],
     'axes.edgecolor': '#bcbcbc',
-    'axes.facecolor': '#eeeeee',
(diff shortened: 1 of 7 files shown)


JaviMerino:
Nice :heart: !  Not only they look consistent, but the static plots are clearer, even the bar plots!

Issue #145, opened 2016-02-25
Parser() should accept a filters parameter


JaviMerino:
`Parser()` should accept a `filters` parameter, like plotter.  Sometimes we need to test behaviours that happen only for some events but not all.  For example, check that a given behaviour happens for a given task:

    prs = Parser(trace, filters={"comm": "task_r5_10-60"})
    prs.solve(statement)

PR #148, opened 2016-02-26
Trigger improvements

JaviMerino:
Allow trigger filters to be any callable and narrow the data frame with the pivot before applying the filter.


sinkap:
Excellent! I am merging this! I would also like to know if we are adding more functionality in the stats framework. I know @derkling was mentioning he intended to add new features. I really like this! Exciting times!

JaviMerino:
We are!  We are adding filters to `Parser`, see #145

PR #150, opened 2016-03-03
ILinePlot: unique x-axis for multiple traces and cleanup

mdigiorgio:
These two commits are:

1. A little cleanup of ILinePlot removing unused attribute
2. Create a unique x-axis when multiple traces with different x-axis values are to be plotted


JaviMerino:
Can you also add a couple of entries in the interactive plotting notebook showcasing this?

JaviMerino:
Merged as #151

PR #153, opened 2016-03-07
Fix parsing special fields

JaviMerino:
@mdigiorgio got a trace with "[" as part of a task name and that breaks trappy.  This series fixes it and also does some spring cleaning on the `test_parse_special_fields` test


sinkap:
Looks good to me!

PR #155, opened 2016-03-08
plotter: ILinePlot: remove undesired horizontal scollbar below plot

### LINKED ISSUE #152: ILinePlot: remove horizontal scrollbar below plots
The interactive plotter shows a horizontal scrollbar below the plot which is not necessary because there is nothing to scroll:

screenshot - 040316 - 12 22 44

mdigiorgio:
When plotting with interactive plotter an unnecessary horizontal scrollbar
appears below the chart. It doesn't actually scroll anything. In fact, it is
enough to remove 2 pixels from the figure to get rid of it.

close #152


mdigiorgio:
This is the simplest way of doing it. If we want something smarter we could play with the div width and compare it to the width of the graph. Which one should we go for?

PR #156, opened 2016-03-15
Fix ILinePlot with duplicated indexes

JaviMerino:


mdigiorgio:
That's good. It looks fine to me, we can then add other tests. One I have in my mind concerns checking this situation:
 - we have two data frames with different indexes
 - ILinePlot will merge them
 - the set of indexes of the merged data frame must be the union of the set of indexes of the initial data frames

So, yes it is a good starting point and we can add things later on :+1:

JaviMerino:
Ok, let's merge it then.  Thanks for the review!

PR #160, opened 2016-03-24
Systrace support

JaviMerino:
This is v2 of #90 .  A couple of commits only needed minor adaptation and I have retained them.  The test also comes from @derkling (thanks!).  With a new `trappy.SysTrace` class, we should now be able to do `trappy.SysTrace("/path/to/trace.html") and from there be able to use all trappy goodies.

### DIFF
--- a/tests/test_systrace.py
+++ b/tests/test_systrace.py
@@ -31,6 +31,10 @@ class TestSystrace(utils_tests.SetupDirectory):
         events = ["sched_switch", "sched_wakeup", "trace_event_clock_sync"]
         trace = trappy.SysTrace("trace.html", events=events)
 
+        self.assertTrue(hasattr(trace, "sched_switch"))
+        self.assertEquals(len(trace.sched_switch.data_frame), 4)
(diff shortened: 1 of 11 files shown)


sinkap:
Thanks for adding this @JaviMerino I am testing this. This looks good to me. Will merge this once I read through the complete patch-set.

derkling:
I've played a bit with this patchset... I :heart: it!
It's our bridge towards the Android word!... so overall, from my side, it deserves a :+1:

There are some minor issues however:
1. the more generic parser for special fields merged with 612384c42c is actually not generic enough to parse systrace generated events.
2. the trace plotter is not yet aware about systrace generated traces

For both these issues I have patches, which I've posted with PR  #162.

With #162 these patches on top of this series I'm able to use systrace to run benchmarks on a N5X target, collect and parse events, plot_trace() these events and get access to all the standard set of trace events we commonly use for EAS development and testing.

Thus, from my side, this is a :+1:  for merge! ;-)

PR #161, opened 2016-03-25
Travis CI support for continuosly running nosetests

mdigiorgio:
This in an initial proposal for using Travis CI to run `nosetests` automatically when something new is pushed to the repository.

*NOTE*: in order to run on master it might be necessary that someone with admin rights on ARM-Software enables TRAPpy on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: python
+python:
+  - "2.7"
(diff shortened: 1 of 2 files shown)


JaviMerino:
This is great, thanks! :heart:

PR #162, opened 2016-03-28
Fixes on top of systrace support

derkling:
The three patches on top of this series fixes some issues discovered while reviewing PR #160.
Thus, this series is to be applied after #160 has been merged.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -106,6 +106,8 @@ class TestBase(utils_tests.SetupDirectory):
                 1004.456789 : { 'task': 'kworker/1:1H',      'pid': 4123, 'cpu': 004 },
                 1005.456789 : { 'task': 'jbd2/sda2-8',       'pid': 5123, 'cpu': 005 },
                 1006.456789 : { 'task': 'IntentService[',    'pid': 6123, 'cpu': 005 },
+                1006.456789 : { 'task': r'/system/bin/.s$_?.u- \a]}c\./ef[.12]*[[l]in]ger',
+                                'pid': 1234, 'cpu': 666 },
         }
 
         in_data = """"""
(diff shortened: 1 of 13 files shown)


tests/test_base.py
```diff
@@ -106,6 +106,8 @@ def test_parse_special_fields(self):
                 1004.456789 : { 'task': 'kworker/1:1H',      'pid': 4123, 'cpu': 004 },
                 1005.456789 : { 'task': 'jbd2/sda2-8',       'pid': 5123, 'cpu': 005 },
                 1006.456789 : { 'task': 'IntentService[',    'pid': 6123, 'cpu': 005 },
+                1006.456789 : { 'task': r'/system/bin/.s$_?.u- \a]}c\./ef[.12]*[[l]in]ger',
+                                'pid': 1234, 'cpu': 666 },
         }
```
JaviMerino: Looks good to me! :+1:

trappy/plotter/__init__.py
```diff
@@ -70,7 +70,10 @@ def plot_trace(trace,
         raise RuntimeError("plot_trace needs ipython environment")
 
     if not isinstance(trace, trappy.BareTrace):
-        trace = trappy.FTrace(trace)
+        if trace.endswith("html"):
+            trace = trappy.SysTrace(trace)
+        else:
+            trace = trappy.FTrace(trace)
```
JaviMerino: I don't like having to rely on file names but in this case I don't think we have an alternative so let's do it :shipit:

trappy/plotter/__init__.py
```diff
@@ -69,7 +69,7 @@ def plot_trace(trace,
     if not IPythonConf.check_ipython():
         raise RuntimeError("plot_trace needs ipython environment")
 
-    if not isinstance(trace, trappy.FTrace):
+    if not isinstance(trace, trappy.BareTrace):
```
JaviMerino: Good catch, thanks!

PR #169, opened 2016-04-01
ftrace: fix default name for traces

### LINKED ISSUE #168: Multitrace multicolumn plots don't display the correct title on top of each graph
See the last plot of `doc/Plotter.ipynb`.  The titles used to be "Trace 0" and "Trace 1".  If you execute it now you get just ".".

trappy_issue

mdigiorgio:
Having "." as default name for traces was preventing TRAPpy from
assigning a meaningful name, like "Trace 0", to traces for which the
user does not explicitly set a name. As a consequence, every plot with
such traces was showing a dot in the legend and in the title (where
trace name is relevant).

close #168

### DIFF
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -476,7 +476,7 @@ class FTrace(GenericFTrace):
 
     """
 
-    def __init__(self, path=".", name=".", normalize_time=True, scope="all",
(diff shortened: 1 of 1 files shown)


JaviMerino:
Oops, my bad, thanks!

PR #170, opened 2016-04-02
README.md: add a link to the API reference

mdigiorgio:


README.md
```diff
@@ -61,3 +61,7 @@ TRAPpy's home directory:
 	$ nosetests
 
 If the installation is correct all tests will succeed.
+
+# API reference
```
JaviMerino: We already have this, there's a "Documentation" heading.  Can you submit a patch with only the "Installaion" typo fixed?

mdigiorgio: Ops, didn't notice that :). Yes, I'll fix that...

Issue #158, opened 2016-03-21
Notebooks should be able to embed data and run stand-alone


sinkap:

sinkap:
This allows ease of sharing notebooks with Interactive plots!

sinkap:
Fixed. Both ILinePlot and EventPlot work standalone now.

PR #178, opened 2016-04-13
Embedded data

sinkap:
Adding Embedded data support for EventPlot. If this looks good I will follow up with an implementation for   ILinePlot

### DIFF
--- a/trappy/plotter/EventPlot.py
+++ b/trappy/plotter/EventPlot.py
@@ -114,14 +114,9 @@ class EventPlot(AbstractDataPlotter):
         graph["keys"] = sorted(keys, key=lambda x: avg[x], reverse=True)
         graph["showSummary"] = summary
         graph["stride"] = AttrConf.EVENT_PLOT_STRIDE
+        self._data = json.dumps(graph)
 
-        json_file = os.path.join(
-            IPythonConf.get_data_path(),
-            self._fig_name +
-            ".json")
 
-        with open(json_file, "w") as json_fh:
-            json.dump(graph, json_fh)
 
         # Initialize the HTML, CSS and JS Components
         self._add_css()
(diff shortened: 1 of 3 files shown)


sinkap:
Issue #158

sinkap:
@derkling Here we go!

sinkap:
After some testing by @mdigiorgio and my discussion with him, we would need to embed the javascript dependencies within the notebook:

d3
EventPlot.js

This is not easily possible. However, I am thinking of using the nbconvert API to write a processor for the notebook which adds these dependencies at conversion time. What are your thoughts on this?

Some reference material for what I am talking about:

http://nbconvert.readthedocs.org/en/latest/nbconvert_library.html

JaviMerino:
Why the config value?  Why don't we always put it in the notebook?  Is there any value in running with `EMBED_DATA = False`?

trappy/plotter/EventPlot.py
```diff
@@ -212,13 +207,28 @@ def _init_html(self):
                         "exports":  "EventPlot"
                     }
                 }
-            });
-            req(["require", "EventPlot"], function() {
-               EventPlot.generate('""" + self._fig_name + """', '""" + IPythonConf.add_web_base("") + """');
-            });
+            });"""
```
JaviMerino: Nit: add a newline here so that `var data = ...` starts in a new line. I.e.:
```
            });
"""
```

trappy/plotter/EventPlot.py
```diff
@@ -212,13 +207,28 @@ def _init_html(self):
...
+
+        generate_call = """
+        req(["require", "EventPlot"], function() {
+            EventPlot.generate('""" + self._fig_name + """', '""" + IPythonConf.add_web_base("") + """', data);
+        });
```
JaviMerino: You don't have to assign it to a variable, just add the string straight away to `div_js` below

PR #179, opened 2016-04-19
ILinePlot: test independent series merging

mdigiorgio:
This tests checks the following:

    - given two data series with different indexes
    - ILinePlot will merge them using _fix_indexes
    - the set of indexes of the merged series must be the union of the set of
      indexes of the initial series


JaviMerino:
I must admit that I hate that series that actually are an overengineered dictionary.   One of this days I'll fix it.

PR #181, opened 2016-04-25
Prettier css

JaviMerino:
When EventPlot embeds the css into the notebook, it adds an extra `\n`
for every line.  `css_fh.readlines()` returns all the lines as an array
but the lines already contain a `\n` at the end of the line.  The
`\n`.join() afterwards adds another `\n`, so you end up with an extra
newline for every line.  Drop the `\n` when reading the css so that it
is embedded without empty lines in between every line in the notebook.


JaviMerino:
I know it doesn't provide any benefit for the user but it annoys me to no end every time I look at the notebook source or the generated html.

PR #182, opened 2016-04-25
Embedded data v2

sinkap:
#158

### DIFF
--- a/trappy/nbexport/templates/trappy.tpl
+++ b/trappy/nbexport/templates/trappy.tpl
@@ -1,0 +1,9 @@
+{% extends 'full.tpl'%}
+{% block html_head %}
+<script src="http://localhost:8080/EventPlot.js" type="text/javascript" charset="utf-8"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.6/d3.min.js", type="text/javascript" charset="utf-8"></script>
+<script src="http://labratrevenge.com/d3-tip/javascripts/d3.tip.v0.6.3.js" type="text/javascript" charset="utf-8"></script>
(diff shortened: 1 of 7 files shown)


trappy/nbexport/templates/trappy.tpl
```diff
@@ -0,0 +1,9 @@
+{% extends 'full.tpl'%}
+{% block html_head %}
+<script src="http://localhost:8080/EventPlot.js" type="text/javascript" charset="utf-8"></script>
```
JaviMerino: `localhost`? on 8080?  I think you forgot to remove this line.

trappy/nbexport/templates/trappy.tpl
```diff
@@ -0,0 +1,9 @@
...
+<script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.6/d3.min.js", type="text/javascript" charset="utf-8"></script>
+<script src="http://labratrevenge.com/d3-tip/javascripts/d3.tip.v0.6.3.js" type="text/javascript" charset="utf-8"></script>
+<script src="https://cdn.rawgit.com/sinkap/a98037133fb79195583a1aa846671c22/raw/c23c231cbb1cdfe29eb84b1f786090a13c16dd93/EventPlot.js" type="text/javascript" charset="utf-8">
```
JaviMerino: This is a bad idea, we now have two copies of `EventPlot.js` that have to be kept in sync.  I'm looking into inlining EventPlot.js to the notebook instead of using `require` to load it.  With that we won't need this line.

sinkap: If you can inline it, it would be great! But this would require you to
change the notebook template. Which, I honestly think is more difficult to
control. It would need the users to use a custom template.

The CDN link can be created when a release is made via the gist API, which
I think should be manageable.

JaviMerino:
Merged as #183

PR #189, opened 2016-05-09
ILinePlot: preserve order in the legend of ILinePlot

JaviMerino:
In aa4bece3adc7 ("ILinePlot: don't call data_frame what's just a
python dict") we turned a pd.Series to a dict, because we were
effectively using it as a dict.  However, there was a behaviour change
as a result of this:  pandas series preserve the order of entries
whereas python dictionaries may reorder its entries.  Use an OrderedDict
instead, as that's what aa4bece3adc7 should have done.

### DIFF
--- a/trappy/plotter/ILinePlot.py
+++ b/trappy/plotter/ILinePlot.py
@@ -19,6 +19,7 @@ classes.  This plot only works when run from an IPython notebook
 
 """
 
+from collections import OrderedDict
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


JaviMerino:
Reported by @deggeman

deggeman:
Tested it. Works again!

sinkap:
@deggeman Thanks for blessing this. @JaviMerino Thanks for the fix. Merging it! :+1:

PR #190, opened 2016-05-09
Document trappy nbexport

JaviMerino:
Document how to use trappy.nbexport to generate HTML versions of notebooks


JaviMerino:
I've followed the boyscout's rule, and I've tried to left the file cleaner than I found it :wink: .  A bit OCDian I must admit.

sinkap:
🉑 ㊗️ 👍

PR #192, opened 2016-06-01
StaticPlot: pass xlim and ylim through to matplotlib

mdigiorgio:
In Static plots, xlim and ylim were ignored causing the user to not see the
effect of setting limits to X and Y axes.

Also, an example of usage of those parameters is added to the Plotter notebook
as a reference on how to use them.


trappy/plotter/StaticPlot.py
```diff
@@ -227,7 +227,7 @@ def _resolve(self, permute, concat):
             num_of_axes,
             width=self._attr["width"],
             length=self._attr["length"],
-            title=self._attr['title'])
+            title=self._attr["title"])
```
JaviMerino: ok, but it's not related to the commit so don't change it.

PR #193, opened 2016-06-06
README.md: Correct apt to apt-get

nuxeh:
Fix readme.


JaviMerino:
I prefer `apt`, it's what Debian recommends for users, whereas `apt-get` is recommended for scripts.  Why do you propose the change?

nuxeh:
Yes, i never realised `apt` was a valid way to install packages, i've always used apt-get, personally.
I'll close this pull request.

PR #195, opened 2016-06-12
systrace: support legacy systrace format

Leo-Yan:
In some old systrace html file it doesn't include the key words for
"" to indicate trace beginning and '<script
class="trace-data"' for trace raw data beginning.

So this patch tries to find compatible string for old and new systrace
format. It changes to use "<title>Android System Trace</title>" and
"  var linuxPerfData" to indicate trace start and trace data start.

### DIFF
--- a/trappy/systrace.py
+++ b/trappy/systrace.py
@@ -30,10 +30,12 @@ the trace
 
     def __call__(self, line):
         if self.before_begin_trace:
-            if line.startswith("<!-- BEGIN TRACE -->"):
+            if line.startswith("<!-- BEGIN TRACE -->") or \
+               line.startswith("<title>Android System Trace</title>"):
                 self.before_begin_trace = False
         elif self.before_script_trace_data:
-            if line.startswith('  <script class="trace-data"'):
+            if line.startswith('  <script class="trace-data"') or \
+               line.startswith("  var linuxPerfData"):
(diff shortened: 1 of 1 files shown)


derkling:
Hi Leo, considering that there are no better version tokens to use, this is perhaps the only suitable solution. Have a couple of questions however:
1. Is there any non trace-event at the end of the trace?
    Otherwise we should better specialize the trace_hasnt_finished() method
2. Can you provide a small example trace to be used for a test?
This allows us at least to track "by example" how a specific trace version is formatted.

This is an example of the format we support at the time:
  https://github.com/ARM-software/trappy/blob/master/tests/trace_systrace.html
and this is the test which consumes that trace:
  https://github.com/ARM-software/trappy/blob/master/tests/test_systrace.py

If you add a test, I'm quite sure you will make Javi much more happy to merge ;-)

JaviMerino:
As @derkling says, I'd like to have a test for this.  Is it possible to add an example trace (reduced and anonymized if needed) so that we can check that we don't break it in the future?

Leo-Yan:
Yes, I will add it and send new patch. Sorry for some delay. :)

JaviMerino:
No need to say sorry, it took me 2 days to have a look at your patch! :smiley:

PR #196, opened 2016-06-13
plotter: Remove handling duplicate indexes from Constraint

JaviMerino:
After 9fe317fdb547 ("plotter: fix constraint filtering with multiple
filters and pivots") it is not necessary to handle duplicate indexes in
Constraint any more as we don't need to reindex dataframes.  Kill this
code from Constarint.

Note that this does not kill handle_duplicate_index from trappy, there
are still valid uses in ILinePlot and the grammar.

### DIFF
--- a/trappy/plotter/Constraint.py
+++ b/trappy/plotter/Constraint.py
@@ -30,7 +30,6 @@ a data column, data event and the requisite filters is
 from trappy.plotter.Utils import decolonize, normalize_list
 from trappy.utils import listify
 from trappy.plotter import AttrConf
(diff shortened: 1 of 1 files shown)


JaviMerino:
@sinkap argued in https://github.com/JaviMerino/trappy/commit/068e89b9e21d9f695e55e15a299c64d92b5f67a9#commitcomment-16321762 that it is needed but having looked at it again, I don't think it is any more.  I vote for removing it, if anything breaks we can always revert this commit and add a test for it.

PR #197, opened 2016-06-16
EventPlot: Don't use a variable for the data

JaviMerino:
When we generate the data for EventPlot, we store it in a variable
called "data".  If there are two plots in a notebook and the notebook is
copied (or the page reloaded), the second "data" variable clobbers the
first one, so when requirejs gets round to execute EventPlot.generate(),
all EventPlots use the same data.

We don't really need to store the data in a variable, we can just pass
it straight to EventPlot.generate()


JaviMerino:
Reported by @derkling .  I'm sure it fixes your issue, not sure if you can test it though.

sinkap:
Ditto! :+1: Vacation delays. :sob:

PR #198, opened 2016-06-16
trappy: set the version using an explicit file

JaviMerino:
There is no pythonic way of specifying the version of a project, [0]
describes 7 (seven!) ways of doing it.  We were currently using method
5, setting the value in setup.py and using pkg_resources to get it from
the installed version.  This works ok if you have installed the package
using "python setup.py" or pip, but fails if you are importing trappy
from a checkout, which is what lisa do.  Even worse, if you import it
from lisa but have an old trappy version installed, trappy.__version__
will tell you the version of the installed trappy, not the one you have
imported and are using.

Switch to use a version.py file that's distributed with the
project (method 3).  trappy.__version__ now reads the imported trappy's
version, which is what you want.  setup.py and the documentation read
the file when they are installing so as to avoid repeating the version
number there.

[0] https://packaging.python.org/en/latest/single_source_version/


sinkap:
Sorry for delay on not being able to get to this, been on vacation. Post facto LGTM from me :+1:

PR #199, opened 2016-06-17
Plotter: color everywhere

JaviMerino:
Since  35e4045 the static plotters (`LinePlot` mainly) support an additional `colors` parameter.  Extend this support and interface to `ILinePlot`.  Add support for it in the signal interface: trace_event:column:color with color a comma-spearated rgb value between 0 and 255.  For example, `thermal:temp:255,0,0` plots the temperature in the thermal trace in red.


JaviMerino:
@jlelli and @derkling asked for this, it would be great if you test this.  Looking forward to colorful plots!

trappy/plotter/AbstractDataPlotter.py
```diff
@@ -73,16 +74,22 @@ def _parse_value(self, signal_def):
         :type signal_def: str
         """
 
-        event, column = signal_def.split(":")
+        match = re.match("(?P<event>[^:]+):(?P<column>[^:]+)(:.+)?", signal_def)
```
derkling: Maybe better to use group names for consistency with other fields?

This should enforce also a check on color format:
```python
match = re.match("r(?P<event>[^:]+):(?P<column>[^:]+)(?P<color>:[0-9]+,[0-9]+,[0-9]+)?", signal_def)
#....
color = match.group("color")
if color:
    color = [int(n) for n in color[1:].split(",")]
else
    color = None
```

doc/Plotter.ipynb
```diff
@@ -123,6 +123,33 @@
      "prompt_number": 4
     },
     {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "Use an additional parameter to specify the color of the line in rgb: \"event_name:column:color\".  The color is specified as a comma separated of rgb values, from 0 to 255.  E.g. 255,0,0 is red and 86,40,32 is brown.  Similar to the example above, to plot the temperature in purple and the previous temperature in green you can do:"
```
derkling: Something like that can be added to plotter::ILinePlot.py to complete the docstring for the "signals" param.

trappy/plotter/AbstractDataPlotter.py
derkling: And what about this simple mod...
```
color = [int(n) if 'x' not in n else int(n, 16) for n in color[1:].split(",")]
```

That should allow to specify colors in HEX format as well, e.g.
```
color = ':0,0xFF,12'
```

derkling:
Few comments but for me this series is OK, awesome addition ;-)
Thanks!

doc/Plotter.ipynb
JaviMerino: Good point, I've added it to `ILinePlot`'s and `LinePlot`'s docstring

trappy/plotter/AbstractDataPlotter.py
JaviMerino: regarding the first suggestion, I have added the group name but dropped the check on color.  With it we can warn the user if they make a mistake.  For example: thermal:temp:23,43.0 silently ignores the color.  I prefer to capture it in the regexp and then barf if we can't parse it.

trappy/plotter/ILinePlotGen.py
```diff
@@ -34,6 +34,18 @@
 from IPython.display import display, HTML
 
 
+def to_dygraph_colors(color_map):
```
sinkap: This function should probably lie in trappy/trappy/plotter/ColorMap.py. The functionality is quite cool and I think the ColorMap.py can be extened to have a set of predefined colors (a mapping of RGB values and the palette should be documented in pydocs. The latter being out of the scope of this pull request. I can probably do some work here.

sinkap:
:+1: from me as well. A great addition.

trappy/plotter/ILinePlotGen.py
JaviMerino: I'm not sure if you wanted it to be part of the ColorMap class or just that the function should be in `ColorMap.py` . I've done the latter because it's too much boilerplate to convert the list of numbers into matplotlib's ScalarMappable only to be converted back into a list of strings with numbers.

JaviMerino:
Merged as #204

PR #200, opened 2016-06-20
systrace: support legacy systrace format

Leo-Yan:
In some old systrace html file it doesn't include the key words for
"" to indicate trace beginning and '<script
class="trace-data"' for trace raw data beginning.

So this patch tries to find compatible string for old and new systrace
format. It changes to use "<title>Android System Trace</title>" and
"  var linuxPerfData" to indicate trace start and trace data start.

### DIFF
--- a/tests/test_systrace.py
+++ b/tests/test_systrace.py
@@ -50,3 +50,38 @@ class TestSystrace(utils_tests.SetupDirectory):
 
         self.assertTrue(hasattr(trace, "_cpus"))
         self.assertEquals(trace._cpus, 3)
+
+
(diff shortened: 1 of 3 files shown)


JaviMerino:
Merged as 459dada.  I only fixed a typo in the class name (`TestLagecySystrace` :arrow_right: `TestLegacySystrace`) and change the docstring of `test_systrace_html` and `test_cpu_counting` so that they are different from their equivalent in `TestSystrace` and we can identify them when we run nosetests with `-v`.

Thanks for the patch @Leo-Yan ! :smile:

Leo-Yan:
Shame for many minor errors. Thanks for fixing and merging.

Issue #103, opened 2015-12-11
Improve Performance of ILinePlot


sinkap:
The performance of ILinePlot currently bottlenecks on the function 

```ILinePlot.convertToDataTable in ILinePlot.js```

![bottleneck](https://cloud.githubusercontent.com/assets/2152812/11745335/61807846-a00f-11e5-8fc3-7977b43d9dbb.png)

We can avoid this overhead by doing this conversion in python.

John-P:
When plotting really large data sets I had some issues even when converted in python. However (this site)[http://kaliatech.github.io/dygraphs-dynamiczooming-example/] has some information about plotting really large datasets.

sinkap:
Thanks @John-P Will give this a read :+1:

JaviMerino:
Fixed by fea80b84fa13a9c0ce75be7fe8c01151f813ee0f

PR #203, opened 2016-06-30
ILinePlot: avoid generating plots that can hang the browser

JaviMerino:
When creating huge plots, dygraph can make the browser choke.  Prevent it by rasing an exception if the user is trying to create a plot that can hang their browser.

### DIFF
--- a/trappy/plotter/ILinePlot.py
+++ b/trappy/plotter/ILinePlot.py
@@ -150,7 +150,12 @@ class ILinePlot(AbstractDataPlotter):
         raise NotImplementedError("Not Available for ILinePlot")
 
     def view(self, test=False):
-        """Displays the graph"""
+        """Displays the graph
+
+        :param test: For testing purposes.  Only set to true if run
+        from the testsuite.
+        :type test: boolean
+        """
 
         # Defer installation of IPython components
         # to the .view call to avoid any errors at
--- a/trappy/plotter/ILinePlot.py
+++ b/trappy/plotter/ILinePlot.py
@@ -149,8 +149,16 @@ class ILinePlot(AbstractDataPlotter):
     def savefig(self, *args, **kwargs):
         raise NotImplementedError("Not Available for ILinePlot")
 
-    def view(self, test=False):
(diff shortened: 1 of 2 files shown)


JaviMerino:
@derkling and @jlelli rightfully complain that having to check whether the trace is too big before plotting is a pain.  It's better to fail if `ILinePlot` is about to plot something that is too big and let the user override it if they are happy to wait.

sinkap:
I am really sad this is needed and I am aware of this problem. I will try to get a handle on #103 as soon as possible since this seems to be bothering us often. Thanks for the fix to let the user be aware of the current limitations.

trappy/plotter/ILinePlot.py
```diff
@@ -149,9 +149,17 @@ def __init__(self, traces, templates=None, **kwargs):
     def savefig(self, *args, **kwargs):
         raise NotImplementedError("Not Available for ILinePlot")
 
-    def view(self, test=False):
+    def view(self, max_datapoints=75000, test=False):
         """Displays the graph
 
+        :param max_datapoints: Maximum number of datapoints to plot.
+        Dygraph can make the browser unresponsive if it tries to plot
+        too many datapoints.  Chrome chokes at around 75000, Firefox
```
sinkap: These numbers could vary, Can we mention the specs of the machine this was tested on as well? So if the user has a beast of a machine, they can calibrate accordingly (although I fear the relationship might be non-linear)

JaviMerino: I've added the processor and browser versions but I doubt it helps much.  The idea is to stop before it may get slow, the user can then decide to increase max_datapoints for the plot or reduce the size of the plot if they're interested in a certain period of time or if they are not worried about the resolution.  I wouldn't expect anybody to fine tune the number, I don't think you gain much.

JaviMerino:
I had forgotten about #103, I thought the problem was in dygraph.  It's good that it's in our hands to improve the performance then!

JaviMerino:
Merged as #205

PR #204, opened 2016-07-04
Plotter color everywhere v2

JaviMerino:
This is v2 of #199.  Changes since v1:

- Added @derkling's suggestion to accept colours specified in hex format in the signals
- @derkling's suggestion of adding a group name for the colour in the signal regexp
- Improved signal's docstring as suggested by @derkling 
- Moved `to_dygraph_colors()` to `trappy/plotter/ColorMap.py` as suggested by @sinkap 

Since 35e4045 the static plotters (`LinePlot` mainly) support an additional `colors` parameter. Extend this support and interface to `ILinePlot`. Add support for it in the signal interface: trace_event:column:color with color a comma-spearated rgb value between 0 and 255. For example, `thermal:temp:255,0,0` plots the temperature in the thermal trace in red.


sinkap:
for the `def to_dygraph_colors` exactly what I had in mind! Thanks! :+1:

PR #205, opened 2016-07-04
ILinePlot: avoid generating plots that can hang the browser

JaviMerino:
This is v2 of #203.  When creating huge plots, ILinePlot can make the browser choke. Prevent it by rasing an exception if the user is trying to create a plot that can hang their browser.


sinkap:
Sorry for the nit and thanks for doing this. LGTM :+1:

PR #206, opened 2016-07-06
ILinePlot: only pass the necessary data when xlim is passed

JaviMerino:
xlim limits the x axis to a given range, but for ILinePlot we pass all the data to dygraph and then let dygraph apply the window.  That means that we include a lot of useless data in the notebook and we lose performance parsing data that will never be plotted.

Improve xlim for ILinePlot so that it only embeds the data relevant for the plot.


sinkap:
Thanks for adding this!

PR #207, opened 2016-07-07
ILinePlot: Performance Improvements for ILinePLot

sinkap:
Instead of converting the DataFrame to JSON and converting to
the dyrgaph format in JS. Send the prepared data as expected by dygraphs
to the JavaScript Library and get rid of convertToDataTable function.

With the current change the following plot works reasonably well on my
laptop (Core i7 2.2GHz, 16GB RAM)

```
columns = ["tick", "tock", "toe"]
df = pd.DataFrame(numpy.random.randn(300000, 3),
columns=columns).cumsum()

trappy.ILinePlot(df, column=columns).view(max_datapoints=2000001)
```

### DIFF
--- a/trappy/plotter/ILinePlotGen.py
+++ b/trappy/plotter/ILinePlotGen.py
@@ -34,6 +34,18 @@ if not IPythonConf.check_ipython():
 
 from IPython.display import display, HTML
 
+def df_to_dygraph(data_frame):
+    result = {}
+
+    values = data_frame.values.tolist()
+    data = [[x] for x in data_frame.index.tolist()]
+
+    for d, val in zip(data, values):
+            d += val
+
+    result["data"] = data
+    result["labels"] = ["index"] + data_frame.columns.tolist()
+    return result
 
 class ILinePlotGen(object):
     """
@@ -210,7 +222,7 @@ class ILinePlotGen(object):
 
         fig_name = self._fig_map[plot_num]
         fig_params = {}
-        fig_params["data"] = OrderedDict((k, v.T.to_dict()) for k, v in data_dict.iteritems())
+        fig_params["data"] = df_to_dygraph(data_dict)
         fig_params["name"] = fig_name
         fig_params["rangesel"] = False
         fig_params["logscale"] = False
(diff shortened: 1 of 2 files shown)


sinkap:
@derkling and @jlelli Can you try this out?

sinkap:
An attempt to fix #103

trappy/plotter/ILinePlotGen.py
```diff
@@ -34,6 +34,18 @@
 
 from IPython.display import display, HTML
 
+def df_to_dygraph(data_frame):
```
JaviMerino: `data_dict` is not always a dataframe, it can be a dictionary.  Try to plot only one column.  In that case, you can't do `data_frame.values` or `data_frame.index`.

The problem is in `ILinePlot::_fix_indexes()`, it can return a data_frame and it shouldn't.  I think that before this patch we should get `_fix_indexes()` to return a dict and make this function deal with dicts instead of dataframes.

mdigiorgio: `merged_df.to_dict()`?

mdigiorgio: Not sure what happens to the order of the labels though (we use `OrderedDict()` in `_plot()`). Maybe we'll have to put it inside an `OrderedDict()`...

JaviMerino: I think we need to keep the `OrderedDict()`, yes

JaviMerino:
While I'm asking for things, can we also have a cute test for `dict_to_dygraph()` :blush:

JaviMerino:
@mdigiorgio has fix ILinePlot::_fix_indexes() to always return an `OrderedDict()` with dicts as values.  @sinkap , can you rebase this on top of master and change `df_to_dygraph()` into a `dict_to_dygraph()` please?

mdigiorgio:
@sinkap the example you wrote works well also on my PC (Core i7@2.00 GHz, 8 GB RAM) :blush:

sinkap:
Will finish this off soon, sorry for the delay :disappointed:

JaviMerino:
Merged as fea80b8 .

PR #208, opened 2016-07-08
plotter: ILinePlot: make _fix_indexes always return a dictionary

mdigiorgio:

### DIFF
--- a/tests/test_plotter.py
+++ b/tests/test_plotter.py
@@ -308,7 +308,7 @@ class TestILinePlotter(unittest.TestCase):
         dfr = pd.DataFrame([0, 1, 2, 3], columns=["a"])
         iplot = trappy.ILinePlot(dfr, column=["a"])
         s = {"s1": s1, "s2": s2}
-        merged = iplot._fix_indexes(s)
+        merged = pd.DataFrame(iplot._fix_indexes(s))
 
         expected_index = index1 + index2
         expected_index.sort()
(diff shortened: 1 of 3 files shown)


tests/test_plotter.py
```diff
@@ -308,7 +308,7 @@ def test_independent_series_merging(self):
         dfr = pd.DataFrame([0, 1, 2, 3], columns=["a"])
         iplot = trappy.ILinePlot(dfr, column=["a"])
         s = {"s1": s1, "s2": s2}
-        merged = iplot._fix_indexes(s)
+        merged = pd.DataFrame(iplot._fix_indexes(s))
```
JaviMerino: Instead of converting it to a DataFrame to then extract the index and convert it back to a list, I'd rather do this check without pandas, by changing the `assertEquals()` below to:
``` python
self.assertEquals(expected_index, sorted(merged["s1"].keys()))
```

trappy/plotter/ILinePlotGen.py
```diff
@@ -210,7 +210,7 @@ def add_plot(self, plot_num, data_dict, title="", test=False):
 
         fig_name = self._fig_map[plot_num]
         fig_params = {}
-        fig_params["data"] = OrderedDict((k, v.T.to_dict()) for k, v in data_dict.iteritems())
+        fig_params["data"] = data_dict
```
JaviMerino: `v.T.to_dict()` is necessary for the for values that are specified as a series instead of a dict.  Try to plot one column of a dataframe.  I'm not sure how to fix this cleanly though.

PR #212, opened 2016-07-14
ILinePlot: Performance Improvements for ILinePlot v2

sinkap:
v2 for #207 and fixing issue #103

### DIFF
--- a/trappy/plotter/ILinePlot.py
+++ b/trappy/plotter/ILinePlot.py
@@ -310,4 +310,4 @@ class ILinePlot(AbstractDataPlotter):
             # default
             merged_df = merged_df.interpolate()
 
-        return OrderedDict(merged_df.to_dict())
+        return merged_df
--- a/tests/test_plotter.py
+++ b/tests/test_plotter.py
@@ -292,6 +292,29 @@ class TestILinePlotter(unittest.TestCase):
         with self.assertRaises(ValueError):
             trappy.ILinePlot([dfr1, dfr2]).view(test=True)
 
+    def test_df_to_dygraph(self):
+        """Test the ILinePlot util function: df_to_dygraph"""
+
+        dfr1 = pd.DataFrame([[1, 2],
+                             [3, 4],
+                             [5, 6]],
+                             index=[0., 1., 2.], columns=["a", "b"])
+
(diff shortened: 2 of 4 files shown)


trappy/plotter/ILinePlot.py
```diff
@@ -310,4 +310,4 @@ def _fix_indexes(self, data_dict):
             # default
             merged_df = merged_df.interpolate()
 
-        return OrderedDict(merged_df.to_dict())
+        return merged_df
```
mdigiorgio: This is not enough... We changed this to return a dictionary because it is used by `_plot` and `_plot_concat`, which operate on dictionaries. So if we always want a DataFrame to be passed to `add_plot` we need to convert the dictionary before that call in both `_plot` and `plot_concat`. The following patch should do the job:

```diff
diff --git a/trappy/plotter/ILinePlot.py b/trappy/plotter/ILinePlot.py
index 07bfbfc..5849d61 100644
--- a/trappy/plotter/ILinePlot.py
+++ b/trappy/plotter/ILinePlot.py
@@ -241,11 +241,9 @@ class ILinePlot(AbstractDataPlotter):
 
             if len(data_dict) > 1:
                 data_dict = self._fix_indexes(data_dict)
-            else:
-                data_dict = OrderedDict(
-                    (k, v.to_dict()) for k, v in data_dict.iteritems()
-                )
-            self._layout.add_plot(plot_index, data_dict, title, test=test)
+
+            data_frame = pd.DataFrame(data_dict)
+            self._layout.add_plot(plot_index, data_frame, title)
             plot_index += 1
 
         self._layout.finish()
@@ -274,11 +272,9 @@ class ILinePlot(AbstractDataPlotter):
 
             if len(data_dict) > 1:
                 data_dict = self._fix_indexes(data_dict)
-            else:
-                data_dict = OrderedDict(
-                    (k, v.to_dict()) for k, v in data_dict.iteritems()
-                )
-            self._layout.add_plot(plot_index, data_dict, title)
+
+            data_frame = pd.DataFrame(data_dict)
+            self._layout.add_plot(plot_index, data_frame, title)
             plot_index += 1
 
         self._layout.finish()
```

JaviMerino:
Merged as fea80b8 .

PR #215, opened 2016-07-19
plotter: EventPlot: add user-defined color map

mdigiorgio:
Give the possibility to specify a mapping between events and custom colours in
the plot.

User can specify colours as both colour name or using its HEX representation.

### DIFF
--- a/trappy/plotter/EventPlot.py
+++ b/trappy/plotter/EventPlot.py
@@ -88,6 +88,25 @@ class EventPlot(AbstractDataPlotter):
 
         :param lanes: The sorted order of lanes
         :type lanes: list
+
+        :param color_map: A mapping between events and colours
+            ::
+                { "<name1>" : "colour1",
+                  .
+                  .
+                  .
+                  "<nameN>" : "colourN"
+                }
+
+            Colour string can be:
+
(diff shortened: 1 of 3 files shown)


trappy/plotter/EventPlot.py
```diff
@@ -87,6 +87,16 @@ class EventPlot(AbstractDataPlotter):
...
         :type lanes: list
+
+        :param color_map: A mapping between events and colours
+            ::
+                { "<name1>" : "colour1",
+                  .
+                  .
+                  .
+                  "<nameN>" : "colourN"
+                }
```
JaviMerino: Please document how you specify colors.  If by name, we could point to a table with names (coming from some HTML documentation?).  For hex, maybe put an example `#ff0000` is red or something like that.

JaviMerino:
Can you add the example to the InteractivePlotter notebook?  Maybe just add it to one of the EventPlot examples.

trappy/plotter/EventPlot.py
mdigiorgio: Sure :smile:

mdigiorgio:
:+1:

Issue #219, opened 2016-08-03
[EventPlot] disappearing events while zooming in


mdigiorgio:
When zooming in a certain region of the plot sometimes some events disappear. The following two pictures show the issue:

- before zoom,
![event_plot](https://cloud.githubusercontent.com/assets/6378106/17364235/f9566240-5976-11e6-8a82-2d2cef295295.png)

- after zooming in once
![event_plot_zoomed](https://cloud.githubusercontent.com/assets/6378106/17364251/091194b6-5977-11e6-8167-3ca2fac8d12c.png)

As you can see all events related to `SurfaceFlinger` task disappeared.

JaviMerino:
Do you have trace for this, or some way of reproducing it?

mdigiorgio:
I pushed a branch on my fork with the data and a notebook to reproduce it: https://github.com/mdigiorgio/trappy/tree/eventplot-tasks_latency

mdigiorgio:
Use the following csv files with the code below...

[data.csv.txt](https://github.com/ARM-software/trappy/files/401221/data.csv.txt)
[data_sf.csv.txt](https://github.com/ARM-software/trappy/files/401222/data_sf.csv.txt)

```python
rendert_df = pd.read_csv('./data.csv')
sf_df = pd.read_csv('./data_sf.csv')

rendert_df.set_index('Time', inplace=True)
sf_df.set_index('Time', inplace=True)

def get_events(task_df, task_name):
    wakeup = task_df[(task_df.curr_state == 'W') & (task_df.next_state == 'A')]
    running = task_df[(task_df.curr_state == 'A')]
    preempt = task_df[(task_df.curr_state.isin(['R', 'R+'])) & (task_df.next_state == 'A')]
    suspend = task_df[(task_df.curr_state == 'D') & (task_df.next_state == 'W')]
    
    t_end = wakeup['t_start'] + wakeup['t_delta']
    w_event = [[start, end, task_name] for start, end in t_end.iteritems()]
    t_end = running['t_start'] + running['t_delta']
    r_event = [[start, end, task_name] for start, end in t_end.iteritems()]
    t_end = preempt['t_start'] + preempt['t_delta']
    p_event = [[start, end, task_name] for start, end in t_end.iteritems()]
    t_end = suspend['t_start'] + suspend['t_delta']
    u_event = [[start, end, task_name] for start, end in t_end.iteritems()]
    
    return r_event, w_event, p_event, u_event

[...]

PR #220, opened 2016-08-20
EventPlot: Add ability to plot simultaneous events

sinkap:
The timeline plot was designed to have one "key" running in one "lane".
Here you have the "queueing" key running in multiple lanes and this
confuses the filtering logic. I imagined the each event to be only
present in one lane at a time, much like a single PID can be only on a
single CPU.

This fix allows the EventPlot to plot concurrent events. Events which
occur simultaneously in multiple lanes

### DIFF
--- a/trappy/plotter/EventPlot.py
+++ b/trappy/plotter/EventPlot.py
@@ -29,6 +29,7 @@ import json
 import os
 from trappy.plotter.AbstractDataPlotter import AbstractDataPlotter
 from trappy.plotter import IPythonConf
+from collections import defaultdict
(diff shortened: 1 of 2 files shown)


sinkap:
An attempt to fix #219

mdigiorgio:
Nice! It works flawlessly with the example I posted on #219 . Thanks for this :+1:

JaviMerino:
Good stuff, thanks @sinkap !  Travis CI fail is not related to this error.  I have run the testsuite myself and it passes.  Merging.

PR #224, opened 2016-09-23
base: Add warning when we're at risk of running out of memory

bjackman:
Pandas doesn't seem to properly handle running out of memory, and segfaults; this gives you no useful error message.

This error probably isn't too rare, because the default Lisa config only provisions 1Gb of memory in the Vagrantfile. That means a 500k line ftrace file can cause a segfault.

I've added a hacky warning in base.py. Do you think this is appropriate? If not, we could just put this in the documentation and hope people spot the connection, and perhaps something like this hacky warning could go in Lisa instead?

### DIFF
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -17,6 +17,20 @@
 
 import re
 import pandas as pd
+import warnings
+
+from resource import getrusage, RUSAGE_SELF
+
+def _get_free_memory_kb():
+    try:
+        with open("/proc/meminfo") as f:
+            memfree_line = [l for l in f.readlines() if "MemFree" in l][0]
+            _, num_kb, _ = memfree_line.split()
+            return int(num_kb)
+    except:
+        # Probably either not running on Linux (no /proc/meminfo), or format has
+        # changed (we didn't find num_kb).
+        return None
 
 def trace_parser_explode_array(string, array_lengths):
     """Explode an array in the trace into individual elements for easy parsing
@@ -158,6 +172,12 @@ class Base(object):
         self.data_array.append(data)
 
     def generate_parsed_data(self):
+
+        # Get a rough idea of how much memory we have to play with
+        kb_free = _get_free_memory_kb()
+        starting_maxrss = getrusage(RUSAGE_SELF).ru_maxrss
+        check_memory_usage = True
+
         for (comm, pid, cpu, data_str) in zip(self.comm_array, self.pid_array,
                                               self.cpu_array, self.data_array):
             data_dict = {"__comm": comm, "__pid": pid, "__cpu": cpu}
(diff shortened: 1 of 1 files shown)


trappy/base.py
```diff
@@ -190,6 +200,16 @@ def create_dataframe(self):
                 expl_val = trace_parser_explode_array(val, trace_arr_lengths)
                 self.data_array[idx] = expl_val
 
+        bytes_free = _get_free_memory_bytes()
+        if bytes_free is not None:
+            # Pandas has been observed to segfault here when running out of
+            # memory. Let's make a very rough guess at whether that's likely and
+            # print a warning if we think it is.
+            dataframe_size_guess = 20 * 1024 * len(self.data_array)
```
JaviMerino: Where does the 20 * 1024 come from?  Every row in the data array uses 20Kb?  Can you add that to the comment?

trappy/base.py
```diff
@@ -190,6 +200,16 @@ def create_dataframe(self):
...
+            if (dataframe_size_guess > 0.8 * bytes_free):
+                logging.warning("TRAPpy: Parsing  large trace may cause errors")
```
JaviMerino: `warnings.warn`

JaviMerino:
500K is not a huge trace.  I guess that any time that I have parsed big traces I have done it from outside of lisa/vagrant.  I am not against merging this, but we should consider adding more memory to vagrant if it is going to complain so easily.

trappy/base.py
bjackman: Yeah, I actually got this by just trying different trace sizes and looking at the overall memory of the process. On reflection I think this is actually more hacky than I can bear so I'm going to do some more thinking, there has to be a better way.

bjackman:
Yep, I suggested we add more memory to the Lisa Vagrantfile, @derkling was more in favour of documenting the issue so that people can add more memory themselves as needed. If I can find a reliable way of detecting this cause of failure (so that users see "you need more memory" instead of just "Segmentation fault (core dumped)", I guess that would be fine.

bjackman:
Pushed a new version, *slightly* less hacky :)

trappy/base.py
```diff
@@ -15,8 +15,23 @@
 
 """Base class to parse trace.dat dumps"""
 
+import logging
```
JaviMerino: You can drop this now.

bjackman:
Removed the logging check. 

I tested it by constructing an FTrace object with a particular trace file on VMs with varying memory sizes. As I mentioned in the commit message I haven't tested on systems without /proc/meminfo but I did try changing the `"/proc/meminfo"` string in `_get_free_memory_kb` to garbage, and nothing untoward happened.

I also wasn't able to trigger a case where the warning gets printed but the memory error doesn't occur; the closest I could get is using a 67Mb trace file on a system with 1118164 bytes memory; the warning gets printed, Pandas doesn't segfault, but instead we seem to get OOM killed.

JaviMerino:
Thanks! :heart:

Issue #210, opened 2016-07-08
Make trappy::Base()'s parse raw traces by default


JaviMerino:
`trappy::Base()` started parsing the normal output of trace-cmd because it includes parsed arrays.  When the `sched_switch` event was introduced, we added raw parsing because for some events we don't want the additional "prettifying" that trace-cmd does and we just want the raw values of the trace.  Right now, parsing the raw trace file needs to be explicitly specified, and is only done for `sched_wakeup`, `sched_wakeup_new` and `sched_switch`.  If you are parsing a trace with `sched_switch` and `cpu_frequency` events, then trappy parses two files: `trace.txt` for `cpu_frequency` and `trace.raw.txt` for `sched_switch`.  Arguably, both events could be parsed from the raw trace.  It's better to switch to parsing raw traces by default and only have trappy create the trace.txt when it's parsing events that need it.

JaviMerino:
Suggested by @derkling

sinkap:
Merged #253 which fixes this. Thanks @credp!

PR #228, opened 2016-10-18
idle: Add parser for cpu_idle events

bjackman:


JaviMerino:
Please add the `trace.dat` instead of the `trace.txt`.  The existing `trace.txt`s make it hard to implement #210 (I had the code but I had to change the existing `trace.txt`s as `trace.dat` and I never got time to do it).  Let's not increase the number of them by having them in binary form.  You will have to annotate `TestCpuIdle` with:

``` python
@unittest.skipUnless(utils_tests.trace_cmd_installed(),
                     "trace-cmd not installed")
```

JaviMerino:
travis says that this breaks the tests because `tests/trace_empty.txt` is no longer "empty".   Can you remove the `cpu_idle` events in there to avoid breaking the testsuite?

tests/test_idle.py
```diff
@@ -0,0 +1,42 @@
...
+
+import os
+import sys
+
+import utils_tests
+import trappy
+
+sys.path.append(os.path.join(utils_tests.TESTS_DIRECTORY, "..", "trappy"))
+
+class BaseTestIdle(utils_tests.SetupDirectory):
```
JaviMerino: No need to create `BaseTestIdle` to be inherited by only one test.  Just move the `__init__` to `TestCpuIdle`.

tests/test_idle.py
```diff
@@ -0,0 +1,42 @@
...
+class TestCpuIdle(BaseTestIdle):
+    def test_get_dataframe(self):
+        df = trappy.FTrace().cpu_idle.data_frame
+
+        exp_states = [2, -1, -1, 1, -1, 0, -1, 2, -1, 0, 0, -1, 0, -1, -1, -1, 2]
+        exp_cpus =   [5,  2,  1, 2,  5, 5,  3, 3,  0, 1, 0,  0, 0 , 0 , 5,  1, 5]
+
+        self.assertEqual(len(df), len(exp_states))
+
+        self.assertTrue((df["state"] == exp_states).all())
```
JaviMerino: If/When this fails, it gives an unhelpful message.  It's better to let unittest do the comparison, something like (untested):

``` python
self.assertEquals(df["state"].tolist(), exp_states)
```

Same for the `cpu_id` column.

tests/test_idle.py
bjackman: Good point, will amend (I just copy pasted this from the sched tests, hence the strangeness)

trappy/idle.py
```diff
@@ -0,0 +1,32 @@
...
+from trappy.dynamic import register_ftrace_parser
+
+class CpuIdle(Base):
+    """Parse cpu_idle"""
+
+    unique_word = "cpu_idle"
+
+    def finalize_object(self):
+        """The trace contains "4294967295" instead of "-1" when exiting an idle
+        state.
```
bjackman: Changing this from a docstring to a comment

bjackman:
Pushed a new version using `assert_series_equal`, unfortunately (?) you also have to explicitly do `assert_index_equal` too, `assert_series_equal` only seems to check the values.

tests/test_idle.py
```diff
@@ -0,0 +1,66 @@
...
+@unittest.skipUnless(utils_tests.trace_cmd_installed(),
+                     "trace-cmd not installed")
+class TestCpuIdle(utils_tests.SetupDirectory):
+    def __init__(self, *args, **kwargs):
+        super(TestCpuIdle, self).__init__(
+            [("trace_idle.dat", "trace.dat")],
+            *args,
+            **kwargs)
+
+    def test_get_dataframe(self):
```
JaviMerino: Add a docstring for the function (just one line, like """cpu_idle events are parsed""").  `nosetests` will use it to describe the tests when running `nosetests -v` or when reporting failures.

JaviMerino:
great, thanks! :heart:

PR #229, opened 2016-11-02
Add fill_alpha param to ILinePlot

bjackman:
Example use case: https://gist.github.com/bjackman/559553d565b033ad3cd7799bf9f9c0de

(Javascript doesn't get run from Gists so you'll need to download it)

### DIFF
--- a/trappy/plotter/ILinePlotGen.py
+++ b/trappy/plotter/ILinePlotGen.py
@@ -235,6 +235,9 @@ class ILinePlotGen(object):
         fig_params["title"] = title
         fig_params["step_plot"] = self._attr["step_plot"]
         fig_params["fill_graph"] = self._attr["fill"]
+        if "fill_alpha" in self._attr:
+            fig_params["fill_alpha"] = self._attr["fill_alpha"]
+            fig_params["fill_graph"] = True
         fig_params["per_line"] = self._attr["per_line"]
         fig_params["height"] = self._attr["height"]
 
--- a/trappy/plotter/js/ILinePlot.js
+++ b/trappy/plotter/js/ILinePlot.js
@@ -133,6 +133,9 @@ var ILinePlot = ( function() {
             dateWindow: t_info.dateWindow
         };
 
+        if (typeof t_info.fill_alpha !== 'undefined')
+            options.fillAlpha = t_info.fill_alpha;
+
         if (typeof colors !== 'undefined')
             options["colors"] = colors;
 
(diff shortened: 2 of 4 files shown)


trappy/plotter/ILinePlotGen.py
```diff
@@ -235,6 +235,9 @@ def add_plot(self, plot_num, data_frame, title="", test=False):
         fig_params["title"] = title
         fig_params["step_plot"] = self._attr["step_plot"]
         fig_params["fill_graph"] = self._attr["fill"]
+        fig_params["fill_alpha"] = self._attr.get("fill_alpha", 0.3)
```
JaviMerino: Why 0.3 for the default value?  Isn't 1.0 a better default?  I know that it is not used, but I want to avoid looking at this in three months time and think "there must be a reason for this magic 0.3"

trappy/plotter/js/ILinePlot.js
```diff
@@ -123,6 +123,7 @@ var ILinePlot = ( function() {
             stepPlot: t_info.step_plot,
             logscale: t_info.logscale,
             fillGraph: t_info.fill_graph,
+            fillAlpha: t_info.fill_alpha,
```
JaviMerino: looks like this only works properly in Chrome.  In Firefox the alpha seems to be the same if I use 0.1 or 0.9.  Anyway, that is a dygraph bug, not trappy.

JaviMerino:
We should document `fill` and `fill_alpha` in `doc/InteractivePlotter`.  It's a shame this features are there and not advertised.

trappy/plotter/js/ILinePlot.js
bjackman: Hmm, it works on my FF 49.0 on Ubuntu 14.04. Perhaps try deleting `~/.local/share/jupyter/nbextensions/` and force-refreshing (`ctrl`+`shift`+`R`)?

bjackman:
Pushed a new version without the magic `0.3` and with a very very simple addition to the doc notebook. Decided not to document the poor-man's stacked graph in the example I linked above due to it being rather hacky but I can add that too if you like it.

JaviMerino:
I have dropped the commit that converts the notebook to v4 format (a076773) because there may still be people using Ubuntu 14.04 that install ipython from apt and that needs the old v3 format. We have kept it on v3 on purpose for a while and  would like to continue doing it for a bit more.

Merged as e62d764645d02c373cfee686ddac3de257aed6e3.  Thanks!

PR #230, opened 2016-11-02
Constraint: Fix __str__ for non-string column keys

bjackman:
Pandas DataFrame columns may be named with any hashable object.

I haven't changed the code for when the data comes from a Trappy trace. I don't think it's necessary (because self.column is always a string - right?) and I haven't figured out how to test it.

### DIFF
--- a/trappy/plotter/Constraint.py
+++ b/trappy/plotter/Constraint.py
@@ -183,7 +183,7 @@ class Constraint(object):
         name = self.get_data_name()
 
         if not self._uses_trappy_trace():
-            return name + ":" + self.column
+            return name + ":" + str(self.column)
 
         return name + ":" + \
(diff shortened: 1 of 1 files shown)


JaviMerino:
A simple test to catch the failure is:

``` python
        dfr = pd.DataFrame({12: [1, 2, 3], 13: [3, 4, 5]})

        constraint = Constraint(dfr, AttrConf.PIVOT, 12, template=None,
                                trace_index=0, filters={}, window=None)

        self.assertEqual(str(constraint), "DataFrame 0:12")
```

In general I'm happy with testing internal APIs.  If at some point they fail because that particular API is no longer used we can get rid of the test.

I have added the test to your commit and merged it.  Thanks!

Issue #233, opened 2017-01-04
Remove legacy from trappy/__init__.py


bjackman:
There's some code marked "For backwards compatibility.  Remove by 2016-12-31" in `trappy/__init__.py`.

So let's remove it?

JaviMerino:
#234 is merged, so closing

bjackman:
Cheers. I was under the impression this would auto-close when the PR was merged.

JaviMerino:
You need to say "fixes" before the issue number: https://github.com/blog/1506-closing-issues-via-pull-requests

I believe there are other magic words.  I think I've closed issues with just "fix" or "close" or something like that.

PR #234, opened 2017-01-04
Update ARM Ltd. Copright notices & remove legacy code

bjackman:
The copyright unit test fails due to the glorious arrival of the new year, so update the necessary copyright notices.

I'm just updating ARM Ltd. ones, there are a couple of Google Inc. that still say 2016 (the unit tests allows this). CC @sinkap 

Also, while grepping for 2016 I came across some "remove-by" notes, so I'm removing them. https://github.com/ARM-software/trappy/issues/233


sinkap:
Looks good to me.

PR #236, opened 2017-01-05
stats/grammar: Raise helpful error when asked to parse absent events

bjackman:
If you use the Parser to access events that Trappy understands, but that are not
present in the trace, you currently get an inscrutable exception when trying to
access `.loc[ self._window[0]:]` on the empty DataFrame in _get_data_frame.

Ideally attempting to parse absent events would just return an empty DataFrame,
but then we don't know what columns it should have. So instead let's just raise
a more helpful error saying that the event is not present.

### DIFF
--- a/tests/test_stats_grammar.py
+++ b/tests/test_stats_grammar.py
@@ -244,3 +244,13 @@ trappy.thermal.Thermal:temp"
         prs = Parser(trace, filters={"cdev_state": 3})
         dfr_res = prs.solve("devfreq_out_power:freq")
         self.assertEquals(len(dfr_res), 1)
+
+    def test_no_events(self):
+        """Test trying to parse absent data"""
+        trace = trappy.FTrace()
+        prs = Parser(trace)
+
+        # cpu_frequency is an event we know how to parse, but it isn't present
+        # in the test trace.
+        self.assertRaisesRegexp(ValueError, "No events found for cpu_frequency",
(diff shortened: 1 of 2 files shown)


bjackman:
Rebased now that https://github.com/ARM-software/trappy/pull/234 is merged

tests/test_stats_grammar.py
```diff
@@ -244,3 +244,13 @@ def test_filtered_parse(self):
...
+
+    def test_no_events(self):
+        """Test trying to parse absent data"""
+        trace = trappy.FTrace()
+        prs = Parser(trace)
+
+        # cpu_frequency is an event we know how to parse, but it isn't present
+        # in the test trace.
+        self.assertRaisesRegexp(ValueError, "No events found for cpu_frequency",
+                                lambda: prs.solve("cpu_frequency:frequency"))
```
JaviMerino: don't use `lambda`, pass the args the arguments to `assertRaisesRegexp`:

``` python
self.assertRaisesRegexp(ValueError, "No events found for cpu_frequency",
                        prs.solve, "cpu_frequency:frequency")
```

or use the context manager:

``` python
with self.assertRaisesRegexp(ValueError, "No events found for cpu_frequency"):
    prs.solve("cpu_frequency:frequency")
```

bjackman: Nice. Will amend.

bjackman:
Pushed a version with the change.

PR #237, opened 2017-01-16
base: Don't error on event field values containing '='

bjackman:
If a trace field value contains a '=', we currently get a ValueError('Too many
values to unpack'). Instead, let's only split on the first '='.

In practice if a field value contains a '=' it's probably because a kernel
developer typo'd a custom event like:

    /* Note missing space between "%d" and "bar" */
    trace_printk("my_broken_event: foo=%dbar=%d", foo, bar)

So I did consider raising an explicit 'malformed event field' error. But this
approach is more flexible in case someone really wanted to trace fields
containing strings with '=' in them.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -74,7 +74,8 @@ class TestBase(utils_tests.SetupDirectory):
 
     def __init__(self, *args, **kwargs):
         super(TestBase, self).__init__(
-             [("../doc/trace.txt", "trace.txt")],
+             [("../doc/trace.txt", "trace.txt"),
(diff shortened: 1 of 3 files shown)


JaviMerino:
It should be easy to create a test for this, can you add it as part of this commit?

bjackman:
Pushed a version with a simple test. I used a .txt trace file because that way it was easy to chop out the relevant bit, is that OK?

JaviMerino:
It's perfect :heart: , thanks!

PR #238, opened 2017-01-17
[RFC] Some clarifications for plotting code

bjackman:
I've been confused a few times while using the plotting code. This is a set of docstrings and cleanups that I made while trying to understand it. I've marked this RFC because:

- it's very likely some of the stuff I've added is wrong or belongs somewhere else (i.e. it needs careful review from people who properly understand the code)
- Some of it is a bit subjective

So I guess consider this low-priority :)


trappy/plotter/Constraint.py
```diff
@@ -276,6 +276,13 @@ def _expand(self):
                    Len[columns] != 1
                    Len[templates] != 1
             )
+
+            TODO  is the following correct?
+            Permute(
+                   Len[traces] != 1
+                   Len[columns] = 1
```
sinkap: Yes, this is correct and should work when the same column is present in all the events (templates). I agree templates was a really bad name for the variable :(

sinkap:
LGTM and thanks for this!

trappy/plotter/Constraint.py
bjackman: Great, I'll remove the "TODO" and amend. I've figured out what `signals` is (I think it's explained in another function) so perhaps I can add a comment for that too :)

bjackman: Actually, shouldn't `templates` have length 1 here? I.e. you have multiple traces; if they're DataFrames that all have a certain column then `columns` should have length 1, and if they're FTraces that all have a certain event then `templates` should have length 1?

PR #245, opened 2017-05-03
ftrace: Improve error messages when failing to parse trace

bjackman:


trappy/ftrace.py
```diff
@@ -181,11 +184,13 @@ def contains_unique_word(line, unique_words=cls_for_unique_word.keys()):
...
                     break
             else:
-                raise ValueError("No unique in {}".format(line))
+                raise FTraceParseError("No unique word in '{}'".format(line))
 
             line = line[:-1]
 
             special_fields_match = special_fields_regexp.match(line)
+            if not special_fields_match:
+                raise FTraceParseError("Coulnd't match special fields in '{}'".format(line))
```
JaviMerino: typo: Couldn't

JaviMerino: I would have liked a test for the "Couldn't match special fields..." error

JaviMerino:
I have added a test and fixed the typo.  It's merged as b1b1c9ec6deb6cb879c683f7ddaebb98d2f016ac.  Out of curiosity, why did you come accross the special_fields_match error?  by using "version" or "cpu" as unique words?  If that's the case, I wonder if instead of raising an exception we should have just added a `continue`.

bjackman:
Thanks a lot for that! I came across the error when trying to parse a trace that was malformed in a way that trace-cmd didn't seem to mind about but that broke TRAPPy rather confusingly. Sorry, I'm hesitant to go into more detail for confidentiality reasons (otherwise I would share the trace).

JaviMerino:
Fair enough.  I was surprised because I was trying to find out something that caused it to fail and all I could think of was either using "version" as unique word or if `trace-cmd` had failed to parse some trace events and have left raw event in the `trace.dat`.

PR #246, opened 2017-05-07
trappy: Improve performance of parser

joelagnel:
This patch reduces the frequency of calling getrusage resulting in a ~3-5%
performance improvement in parsing. We do a check in the beginning and then
check once every 10000 events.

### DIFF
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -174,9 +174,11 @@ class Base(object):
     def generate_parsed_data(self):
 
         # Get a rough idea of how much memory we have to play with
+        CHECK_MEM_COUNT = 10000
         kb_free = _get_free_memory_kb()
         starting_maxrss = getrusage(RUSAGE_SELF).ru_maxrss
         check_memory_usage = True
+        check_memory_count = 1
 
         for (comm, pid, cpu, data_str) in zip(self.comm_array, self.pid_array,
                                               self.cpu_array, self.data_array):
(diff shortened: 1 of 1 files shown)


trappy/base.py
```diff
@@ -200,13 +202,20 @@ def generate_parsed_data(self):
             # When running out of memory, Pandas has been observed to segfault
             # rather than throwing a proper Python error.
             # Look at how much memory our process is using and warn if we seem
-            # to be getting close to the system's limit.
+            # to be getting close to the system's limit, check it only once
+            # in the beginning and then every CHECK_MEM_COUNT events
+            check_memory_count -= 1
+            if check_memory_count != 0:
+                yield data_dict
+                continue
```
bjackman: I think it would be clearer if you inverted this `if` and partially combined it with the one below, like:
```
check_memory_count -= 1
if check_memory_usage and check_memory_count == 0:
    kb_used = (getrusage(RUSAGE_SELF).ru_maxrss - starting_maxrss)
    if kb_free and kb_used > kb_free * 0.9:
        warnings.warn("TRAPpy: Appear to be low on memory. "
                               "If errors arise, try providing more RAM")
        check_memory_usage = False
    check_memory_count = CHECK_MEM_COUNT
```

joelagnel:
Ping?

JaviMerino:
My only worry is how to come up with a good value for `CHECK_MEM_COUNT`.  10000 is as good as any value, so let's wait for somebody to complain and then we will adjust it.

PR #247, opened 2017-05-15
trappy: Speed up trappy by caching trace parsing

joelagnel:
Pandas is extremely fast at parsing csv to data frames. Astonishingly it takes
< 1s to serialize/deserialize a 100MB work of traces with 430000 events to/from
csv. We leverage this and write out a data frames into a csv file when they are
created for the first time. Next time we read it out if it exists. To make
sure, the cache isn't stale, we take the md5sum of the trace file and also
ensure all CSVs exist before reading from the cache. I get a speed up of 16s to
1s when parsing a 100MB trace.


joelagnel:
CC @bjackman

trappy/ftrace.py
```diff
@@ -43,6 +44,37 @@ def _plot_freq_hists(allfreqs, what, axis, title):
         trappy.plot_utils.plot_hist(allfreqs[actor], ax, this_title, "KHz", 20,
                              "Frequency", xlim, "default")
 
+def _trace_cache_path(trace_file):
+    cache_dir  = '.' +  os.path.basename(trace_file) + '.cache'
+    cache_path = os.path.dirname(trace_file) +  '/' + cache_dir
```
bjackman: Shoud use `os.path.join`

trappy/ftrace.py
```diff
@@ -43,6 +44,37 @@ def _plot_freq_hists(allfreqs, what, axis, title):
...
+    cache_path = _trace_cache_path(trace_file)
+    md5file = cache_path + '/md5sum'
+
+    if os.path.exists(cache_path):
+        shutil.rmtree(cache_path)
+    os.mkdir(cache_path)
+
+    md5sum = hashlib.md5(open(trace_file, 'rb').read()).hexdigest()
+    open(md5file, 'w').write(md5sum)
+
```
bjackman: How come these aren't methods of `GenericFTrace` or `BareTrace`?

trappy/ftrace.py
```diff
@@ -76,6 +108,22 @@ def __init__(self, name="", normalize_time=True, scope="all",
...
 
+        cache_path = _trace_cache_path(self.trace_path)
+        cache_good = False
+
+        if _check_trace_cache(self.trace_path):
+            cache_good = True
+            try:
+                # Read csv into frames
+                for trace_class in self.trace_classes:
+                    csv_file = cache_path + '/' + trace_class.unique_word[:-1] + '.csv'
```
bjackman: - `os.path.join`
- I'm not too comfortable with the use of `unique_word[:-1]` - I wonder if we could just us `trace_class.__name__`?

trappy/ftrace.py
```diff
@@ -76,6 +108,22 @@ def __init__(self, name="", normalize_time=True, scope="all",
...
+                    trace_class.read_csv(csv_file)
+            except:
```
bjackman: Maybe we should log a warning here? Would be annoying if we later broke this code and didn't notice for ages because an `AttributeError` was getting thrown away or something.

trappy/ftrace.py
```diff
@@ -76,6 +108,22 @@ def __init__(self, name="", normalize_time=True, scope="all",
...
+                cache_good = False
+
+        if cache_good:
+            return
```
bjackman: If I'm not mistaken you can get rid of `cache_good`:
```
if _check_trace_cache(self.trace_path):
    try:
            csv_file = cache_path + '/' + trace_class.unique_word[:-1] + '.csv'
            trace_class.read_csv(csv_file)
        return
    except:
        pass
````

Also what if we e.g. successfully parse the CSV for trace class A (so its `data_frame` member is populated), then fail while parsing the CSV for class B? It looks like all the events for A will be duplicated in `__parse_trace_file` below.

File: trappy/ftrace.py [hunk 6]
```diff
@@ -85,6 +133,10 @@ def __init__(self, name="", normalize_time=True, scope="all",
         if normalize_time:
             self.normalize_time()
 
+        _create_trace_cache(self.trace_path)
+        for trace_class in self.trace_classes:
+            trace_class.write_csv(cache_path + '/' + trace_class.unique_word[:-1] + '.csv')
```
bjackman: - `os.path.join`
- The creation of the CSV path is duplicated here, should probably put it in a method

trappy/ftrace.py
joelagnel: I felt these are just utility methods, that didn't as such belong to these classes so I'm just as happy keeping them in the module than within classes

trappy/ftrace.py
joelagnel: yes, great idea. I'll try to use the class name. The only issue is if the class name changes for some reason. But I guess that's rare.

trappy/ftrace.py
joelagnel: ok sure

trappy/ftrace.py
joelagnel: Ok.

trappy/ftrace.py
joelagnel: duplicated how?
data_frame will just be overwritten in that case
https://github.com/joelagnel/trappy/blob/97a29442e2c19bc5c09f73bee83f08d35406c907/trappy/base.py#L225
So I don't see how this is an issue

trappy/ftrace.py
joelagnel: yeah will check why :(

trappy/ftrace.py
joelagnel: I think I can put them in GenericFTrace, that'll be fine.

trappy/ftrace.py
bjackman: Ahhh yes you're quite right, I misunderstood the code.

trappy/ftrace.py
bjackman: Yeah I don't think it's too much of a problem if the cache gets "invalidated" (especially since the cache parsing could theoretically change anyway). That actually got me thinking maybe we should also add a ".cache_version" in th cache directory. Then we could write "0" into that file and check it matches when reading back, so that if we decide to change the layout/format then we can just change that 0 and no need to worry about compatibility? Worth the effort?

trappy/ftrace.py
bjackman: Cool, doesn't really matter was just curious.

trappy/ftrace.py
```diff
@@ -21,6 +21,7 @@
 import os
 import re
 import pandas as pd
+import hashlib, shutil
```
JaviMerino: One import per line please.

trappy/ftrace.py
JaviMerino: Also, please use `os.path.join()`

trappy/ftrace.py
```diff
@@ -43,6 +44,37 @@ def _plot_freq_hists(allfreqs, what, axis, title):
...
                              "Frequency", xlim, "default")
 
+def _trace_cache_path(trace_file):
+    cache_dir  = '.' +  os.path.basename(trace_file) + '.cache'
+    cache_path = os.path.dirname(trace_file) +  '/' + cache_dir
+    return cache_path
+
+def _check_trace_cache(trace_file):
+    cache_path = _trace_cache_path(trace_file)
+    md5file = cache_path + '/md5sum'
```
JaviMerino: `os.path.join()`

joelagnel:
All great suggestions. Will fix up.

About pickle, I still prefer csv because it's easier to investigate the cache in clear text and also is more immune to the internal format of pandas objects than binary object representation I think. Also I am also not sure how much faster we can get because we are already so fast with csv.

joelagnel:
Also from this link it looks like csv is faster than pickle:
http://matthewrocklin.com/blog/work/2015/03/16/Fast-Serialization

trappy/ftrace.py
joelagnel: found why it broke them, because I wasn't using abspath causing mkdir to fail, fixed it in next rev.

trappy/ftrace.py
joelagnel: Actually, I didn't agree with the last point why this should be a try/except? If we fail to write to the cache for some reason, I'd rather treat it as a failure and consider it a bug. Are you worried about the system running out of space?

joelagnel:
So this is the new version with all review comments addressed.

tests/test_caching.py
```diff
@@ -0,0 +1,52 @@
+#    Copyright 2015-2017 ARM Limited
```
bjackman: I'm not a lawyer but this doesn't look right to me :stuck_out_tongue:

trappy/ftrace.py
```diff
@@ -79,6 +120,22 @@ def __init__(self, name="", normalize_time=True, scope="all",
...
+        if not self.__class__.disable_cache and self._check_trace_cache():
+            cache_good = True
+            try:
+                # Read csv into frames
+                for trace_class in self.trace_classes:
+                    csv_file = self._get_csv_path(trace_class)
+                    trace_class.read_csv(csv_file)
+            except:
+                warnings.warn("TRAPpy: Caching disabled due to cache parsing error")
+                cache_good = False
```
bjackman: You can still get rid of `cache_good` ([previous comment here](https://github.com/ARM-software/trappy/pull/247#discussion_r116565256))

trappy/ftrace.py
```diff
@@ -88,6 +145,13 @@ def __init__(self, name="", normalize_time=True, scope="all",
         if normalize_time:
             self.normalize_time()
 
+        if self.__class__.disable_cache: return
+
+        self._create_trace_cache()
+        for trace_class in self.trace_classes:
+            csv_file = self._get_csv_path(trace_class)
+            trace_class.write_csv(csv_file)
+
```
bjackman: [Continuing from [previous thread](https://github.com/ARM-software/trappy/pull/247#discussion_r118583125), which Github won't let me reply to any more]

> why this should be a try/except? If we fail to write to the cache for some reason, I'd rather treat it as a failure and consider it a bug. Are you worried about the system running out of space?

I was actually thinking of something like a permissions error - e.g. sometimes at ARM people share files via NFS mounts where only they have write permissions. Just in general the fact that you intuitively expect to be able to parse traces from read-only places. 

Maybe just catch `OSError`s?

bjackman: That actually got me thinking - would Trappy be a bad OS citizen by dropping these ".cache" directories all around people's filesystems? E.g. on Linux should we use `/var/cache`? That would introduce platform-specific stuff so maybe it's not worth the hassle.

tests/test_caching.py
```diff
@@ -0,0 +1,52 @@
...
+        """Test that cache should not be created when disabled """
+        GenericFTrace.disable_cache = True
+        trace = trappy.FTrace()
+
+        trace_path = os.path.abspath(trace.trace_path)
+        trace_dir = os.path.dirname(trace_path)
+        trace_file = os.path.basename(trace_path)
+        cache_dir = '.' + trace_file + '.cache'
+
+        self.assertFalse(cache_dir in os.listdir(trace_dir))
```
bjackman: Would be good to also test
- That the data parsed from a valid cache is exactly the same as the data from the corresponding trace (I've noticed it's quite easy to do things like mess up column names when doing `DataFrame`<->`CSV`).
- That the checksum test really works, i.e. that you don't get bogus data when there's an invalid cache.

trappy/ftrace.py
joelagnel: Yes, I think management of that cache directory (configuring its path etc) would be a pain, I'd much rather just drop them into the current directory. There csvs are also much smaller than the actual trace so its not a big space consumer.

joelagnel: Ok, I'll catch OSError and print a warning in this case instead of just failing.

trappy/ftrace.py
joelagnel: Yes, sorry will do.

tests/test_caching.py
joelagnel: Hehe will fix.. infact is it Ok to make this "ARM Limited, Google and Contributors" even?

tests/test_caching.py
joelagnel: I'll add these , could you please share what troubles you had when converting `CSV` to `DataFrame` and vice versa?

tests/test_caching.py
bjackman: That sounds fine to my inexpert ears - actually it looks like the [tests](https://github.com/ARM-software/trappy/blob/master/tests/test_copyright.py#L44) require "ARM Limited" to be in the copyright line so you'll need to change that test if you just want to mark something as solely Google copyright (which would also be fine IIUC).

tests/test_caching.py
bjackman: Sorry, I can't actually remember :confounded: I just had a post-it note scribbled on the inside of my brain that said "double check `pd.read_csv` did what you thought it did".

tests/test_caching.py
joelagnel: ah ok, there's even a Google test just after that that checks
https://github.com/ARM-software/trappy/blob/master/tests/test_copyright.py#L48
So probably some files already have Google in their copyrights, I'll double check

tests/test_caching.py
joelagnel: I added the test for data but I'm afraid for the check sum test I'm a bit worn out (I worked every night last few nights on this) - appreciate any help if you want to add more test cases. I think for now the tests I have which check the cached/uncached data are the same, the column names match, and whether caching can be disabled give most of the coverage. I am sure more things can be creatively added here.

joelagnel:
Hopefully this one is merge-ready

tests/test_caching.py
bjackman: Yeah this look good to me, thanks for adding it, I think it's really useful to have coverage like this.

bjackman:
I pushed an extra test, along with some microscopic fixes for stuff I noticed while writing it, to the branch on my repo here: https://github.com/bjackman/trappy/tree/for-trappy-cache

So I guess feel free to cherry-pick/squash that into your branch? Or you can leave it and I'll submit it separately once this is merged (I'm ill this week and cannot think clearly at all so the code may be nonsense).

joelagnel:
Squashed bjackman@'s fixups except the new test case (which IMO can be added after this set is merged). Pushed a new version. @JaviMerino does it look good to you?

bjackman:
This is good from my point of view, let's see what others think.

bjackman:
I updated my branch (https://github.com/bjackman/trappy/tree/for-trappy-cache) with the missing file for the test I added. I also added another test for something I was confused about, and an extra check in _check_trace_cache, to avoid a warning if the set of parsed events has changed since the cache was written.

Also, AFAICT this currently breaks the `normalize_time` and `window`/`abs_window` params. Probably the simplest solution is just to add them to the md5sum or dump them another file or something?

joelagnel:
CC @sinkap

joelagnel:
Updated with @bjackman 's latest commits. Thanks!
Will work on the other corner cases now (new event passed after cache creation). @bjackman while I do that could fix the normalize time issue you mentioned if you had time? I am new to that code so I could use some help there. Thanks!

joelagnel:
@bjackman so I dropped "ftrace: Add explicit check for presence of each .csv file" and added a new commit which partially reads the cache. Your new test case for dynamic events added after caching passes with that.

bjackman:
I updated my branch (https://github.com/bjackman/trappy/tree/for-trappy-cache) with

- A couple more minor fixups I noticed
- A couple more simple tests (which fail without the following commits)
- A commit to re-order the caching operations, to fix normalize_time (and incidentally clean up the impact on `finalize_objects`)
- A commit to store and check the `window` parameter.

sinkap:
@bjackman I will review your branch today. Since it seems to have some more fixes. @joelagnel

joelagnel:
@bjackman I think you didn't update your branch (last update was June 8 to https://github.com/bjackman/trappy/tree/for-trappy-cache and I already picked up those fixes). Also could you rebase and master when you push again? Thanks!

bjackman:
Eek.. you're right I didn't push, and I can't find the commits on my desktop... I will have to check when I get to my laptop :fearful:

sinkap:
@bjackman feel free to ping when you have the updated commits.

joelagnel:
@bjackman not having caching is killing me, can't wait to have this goodness in with your fixes ;D

credp:
I think @joelagnel and @bjackman  addressed Javi's change request, so I am going to dismiss that review to remove the big red cross :+1:

joelagnel:
thanks @credp

credp:
So where are we with this PR? Looks just about ready pending a push from @bjackman? (cc @joelagnel )

sinkap:
Yep. we have a sync meeting today. Let's finalize it there.

bjackman:
Hey guys, really sorry for causing delays here, I forgot about it then went on holiday. Will update tonight.

sinkap:
@bjackman Enjoy your holiday. I am merging #253. It would be best to rebase this :)

bjackman:
OK, I've pushed the commits now: https://github.com/bjackman/trappy/tree/for-trappy-cache

Forgotten the details so I'll do a self-review tomorrow :)

joelagnel:
@bjackman, also could you rebase on top of @credp 's changes which got
merged? :) thanks man.

bjackman:
OK I updated the branch. The rebase was pretty awkward so I think the patch stack is messed up, but TBH I think we should squash them all anyway instead of having "add X", "fix X" in a single PR.

joelagnel:
Thanks @bjackman , squashed as many as that made sense to squash. HAPpy SNAPpy TRAPpy!

sinkap:
Merged!

PR #248, opened 2017-06-03
[RFC] trappy: Add tgid information from systrace

joelagnel:
Add tgid information if available during parsing traces

Change-Id: I7d280a9dc95d48a2730983e5ddb474d266ebf0fd

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -85,7 +85,7 @@ class TestBase(utils_tests.SetupDirectory):
         in_data = """     kworker/4:1-397   [004]   720.741315: thermal_power_cpu_get: cpus=000000f0 freq=1900000 raw_cpu_power=1259 load={} power=61
      kworker/4:1-397   [004]   720.741349: thermal_power_cpu_get: cpus=0000000f freq=1400000 raw_cpu_power=189 load={} power=14"""
 
-        expected_columns = set(["__comm", "__pid", "__cpu", "cpus", "freq",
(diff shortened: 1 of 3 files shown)


joelagnel:
Does anyone know why the stupid CI fails? The CI logs don't make sense to me. I ran nosetests and it runs fine locally for me.

bjackman:
I can't repro the issue either - I noticed that I'm running Python 2.7.6 but the CI is on 2.7.3 - maybe a bug in `re`? I didn't find anything relevant in the Python changelog but [this StackOverflow question suggests there was indeed a bug](https://stackoverflow.com/questions/3675144/regex-error-nothing-to-repeat).

If I'm right, my initial thought was to just update the CI to use a recent Python (I think we could just change the Travis setup to use a more recent Ubuntu) but actually I suppose we should work around this so people can use Trappy on old systems...

bjackman:
I haven't read this yet but I guess this is based on your upstream kernel posting? If so do we need to wait for that to get merged?

joelagnel:
Hi Brendan, We don't have to wait because an older implementation of that
has been in Android kernels since years.

bjackman:
Oh OK gotcha!

bjackman:
Aside from the Python regex bug, there are some places in the tests where we assert on the columns of the DataFrames, so presumably they'll need fixing. Also maybe there's some documentation somewhere that lists the special field columns? I can't see it anywhere though.

joelagnel:
I closed this for now since I've to rewrite it to be completely different and don't want to confuse reviewers.

PR #249, opened 2017-06-12
systrace parsing new edition

joelagnel:
Hi @derkling ,
I went ahead and re-wrote #243. This is very fresh so I am just posting it early for any comments. This approach is more cleaner and also doesn't have issues with #243. It also doesn't require dropping custom strings written by tracing_mark_write that systrace doesn't recognize . I didn't add your test case but I ran some scripts to see it working and existing nosetests are passing.  I would appreciate if you can integrate your test case from #243 to work with this PR and feel free to repost this PR with that. We can also close that old PR and work with this one. I borrowed your regex from the systrace format patch and given you credit in the relevant patch, thanks. CC @bjackman

### DIFF
--- a/trappy/bare_trace.py
+++ b/trappy/bare_trace.py
@@ -134,5 +134,9 @@ class BareTrace(object):
 
     def finalize_objects(self):
         for trace_class in self.trace_classes:
+            trace_class.tracer = self
             trace_class.create_dataframe()
             trace_class.finalize_object()
+
+    def generate_data_dict(self):
+        return None
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -99,6 +99,7 @@ class Base(object):
 
     """
     def __init__(self, parse_raw=False):
+        self.tracer = None
         self.data_frame = pd.DataFrame()
         self.data_array = []
         self.time_array = []
@@ -171,6 +172,25 @@ class Base(object):
         self.cpu_array.append(cpu)
         self.data_array.append(data)
 
+    def generate_data_dict(self, data_str):
+        data_dict = {}
+        prev_key = None
+        for field in data_str.split():
+            if "=" not in field:
+                # Concatenation is supported only for "string" values
+                if type(data_dict[prev_key]) is not str:
+                    continue
+                data_dict[prev_key] += ' ' + field
+                continue
+            (key, value) = field.split('=', 1)
+            try:
+                value = int(value)
+            except ValueError:
+                pass
+            data_dict[key] = value
+            prev_key = key
+        return data_dict
+
     def generate_parsed_data(self):
 
         # Get a rough idea of how much memory we have to play with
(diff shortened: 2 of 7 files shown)


bjackman:
If we add a specific parser class for `tracing_mark_write`, does that break

bjackman:
I don't grok the parsing code but this seems good. My only concern (aside from the little comments on the `tracer` attribute) is this: If we add a parser class for "tracing_mark_write" do we break injecting custom events from userspace (i.e. not systrace)? For example in LISA we use devlib to inject "cpu_frequency_devlib" events, which are really "tracing_mark_write" events containing the string "cpu_frequency_devlib:". We pass "cpu_frequency_devlib" in the `events` parameter to `FTrace` which results in a dynamic parser class being created with that as its unique word. Would that be broken if there was already a parser class with "tracing_mark_write" as its unique word?

joelagnel:
Brendan you closed this PR, hope that wasn't intentional? ;)

You are right this will break that. It could be easily fixed though by making sure tracing mark write class is used only when nothing else matched. I will work on fixing that. I will also add better commit message for explaining the parsing.

bjackman:
> Brendan you closed this PR, hope that wasn't intentional? ;)

Yep, accident, sorry!

trappy/tracing_mark_write.py
```diff
@@ -0,0 +1,33 @@
+#    Copyright 2017 ARM Limited, Google and contributors
```
sinkap: Let's add a module docstring

trappy/tracing_mark_write.py
```diff
@@ -0,0 +1,33 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from trappy.base import Base
+from trappy.dynamic import register_ftrace_parser
+
+class TracingMarkWrite(Base):
```
sinkap: Please add a docstring for the class. I know we have not been careful about doing this. I will probably need to backfill.

trappy/systrace.py
```diff
@@ -14,6 +14,7 @@
 #
 
 from trappy.ftrace import GenericFTrace
+import re
 
 class drop_before_trace(object):
```
sinkap: we have been naming the classes as DropBeforeTrace. Also I am not sure what the name conveys. I will change that in a follow up PR.

trappy/systrace.py
```diff
@@ -75,3 +76,19 @@ def trace_hasnt_finished(self):
...
+        match = SYSTRACE_EVENT.match(data_str)
+        if match:
+            data_dict = { 'event': match.group('event'),
+                          'pid'  : match.group('pid'),
+                          'func' : match.group('func'),
+                          'data' : match.group('data') }
+
+        return data_dict
+
+SYSTRACE_EVENT = re.compile(
```
sinkap: I would move this to the top

trappy/base.py
```diff
@@ -98,7 +98,8 @@ class Base(object):
     This class acts as a base class for all TRAPpy events
 
     """
-    def __init__(self, parse_raw=False):
+    def __init__(self, parse_raw=False, is_backup=False):
```
sinkap: Can you add to the documentation what `is_backup` actually means? Let's call it `is_fallback` as discussed.

trappy/bare_trace.py
```diff
@@ -134,5 +134,9 @@ def add_parsed_event(self, name, dfr, pivot=None):
 
     def finalize_objects(self):
         for trace_class in self.trace_classes:
+            trace_class.tracer = self
             trace_class.create_dataframe()
             trace_class.finalize_object()
+
+    def generate_data_dict(self):
+        return None
```
joelagnel: If we remove it, it will break code that calls it (since not all tracers are expected to implement it). Also not implementing it is not an exception but the normal fall back.

trappy/tracing_mark_write.py
joelagnel: Ok

trappy/systrace.py
joelagnel: Ok

trappy/systrace.py
joelagnel: done

trappy/bare_trace.py
sinkap: Then we should have this raise the exception and we should catch the exception and then implement the fallback. So that it can be more explicit

sinkap:
Okay. I am merging this one. Thanks everyone!

joelagnel:
Merged right?

PR #251, opened 2017-06-13
trappy: Add a line number column

joelagnel:
Useful for joining DataFrames that have timestamp collisions or for iterating
through a group of DataFrames in line order.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -85,7 +85,7 @@ class TestBase(utils_tests.SetupDirectory):
         in_data = """     kworker/4:1-397   [004]   720.741315: thermal_power_cpu_get: cpus=000000f0 freq=1900000 raw_cpu_power=1259 load={} power=61
      kworker/4:1-397   [004]   720.741349: thermal_power_cpu_get: cpus=0000000f freq=1400000 raw_cpu_power=189 load={} power=14"""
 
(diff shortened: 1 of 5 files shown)


joelagnel:
CC @bjackman @sinkap @derkling @bjackman @JaviMerino

bjackman:
Looks good to me :+1:

sinkap:
I will test this and the other PRs today and merge these.

joelagnel:
Hi Patrick,
I will add the systrace test.
Also about storing of exact position in trace file (file offset) will need yet another column because in systrace there is an offset between start of the trace and start of the trace file. I think for now it's ok to just use exact event position from start of the trace and a future patch can store an "offset" field in the FTrace object. Will refresh this PR shortly and then @sinkap can further review, thanks.

derkling:
@joelagnel  the implementation I've provided in here:
https://github.com/ARM-software/lisa/pulls
should already skip everything is not a trace event, i.e. the first trace event will get ID=0.

This is done using a feature already there in trappy which allows to skip non trace event lines... have a try, I think it should work for systrace as well. Meaning the first systrace event you have should get ID=0.

joelagnel:
Hi @derkling yes what I meant is I will use those commits from your PR which do what you wanted, thanks.

sinkap:
merged.

PR #252, opened 2017-06-15
systrace: line test and regex fix

joelagnel:
I promised @derkling I would write this so here you go.

### DIFF
--- a/tests/test_systrace.py
+++ b/tests/test_systrace.py
@@ -64,6 +64,15 @@ class TestSystrace(utils_tests.SetupDirectory):
         self.assertTrue(dfr['func'].iloc[2], 'notifyFramePending')
         self.assertTrue(dfr['data'].iloc[-2], 'HW_VSYNC_0')
 
+    def test_systrace_line_num(self):
+        """Test for line numbers in a systrace"""
+        trace = trappy.SysTrace("trace_sf.html")
+        dfr = trace.sched_switch.data_frame
+        self.assertEquals(trace.lines, 2506)
+        self.assertEquals(dfr['__line'].iloc[0], 0)
+        self.assertEquals(dfr['__line'].iloc[1], 6)
+        self.assertEquals(dfr['__line'].iloc[-1], 2505)
+
 class TestLegacySystrace(utils_tests.SetupDirectory):
 
     def __init__(self, *args, **kwargs):
--- a/tests/test_systrace.py
+++ b/tests/test_systrace.py
@@ -17,6 +17,8 @@ import utils_tests
 
 import trappy
 
+import numpy as np
+
 class TestSystrace(utils_tests.SetupDirectory):
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


joelagnel:
@derkling @bjackman @sinkap looks good to you?

bjackman:
Yep LGTM

tests/test_systrace.py
```diff
@@ -64,6 +64,14 @@ def test_systrace_userspace(self):
         self.assertTrue(dfr['func'].iloc[2], 'notifyFramePending')
         self.assertTrue(dfr['data'].iloc[-2], 'HW_VSYNC_0')
 
+    def test_systrace_line_num(self):
+        """Test for line numbers in a systrace"""
+        trace = trappy.SysTrace("trace_sf.html")
+        dfr = trace.sched_switch.data_frame
+        self.assertEquals(dfr['__line'].iloc[0], 0)
+        self.assertEquals(dfr['__line'].iloc[1], 6)
+        self.assertEquals(dfr['__line'].iloc[-1], trace.lines - 1)
```
derkling: Perhaps it's more strict to check for an hardcoded number.

Indeed the values for `__line` are set out of the value of `trace.lines` and since the `sched_switch` is the last event in the test trace I would expect it always matches. However, this is not checking if we should screw-up on updating `trace.lines`, isn't it?

joelagnel:
Threw in the regex fix here too

joelagnel:
Updated with few more fixes:
* Parsing issue with systrace for 'data' column
* Fixed initial systrace test case

trappy/systrace.py
```diff
@@ -88,9 +88,12 @@ def generate_data_dict(self, data_str):
...
         match = SYSTRACE_EVENT.match(data_str)
         if match:
+            pid = match.group('pid')
+            if pid:
+                pid = int(pid)
             data_dict = { 'event': match.group('event'),
-                          'pid'  : match.group('pid'),
-                          'func' : match.group('func'),
-                          'data' : match.group('data') }
+                          'pid'  : pid                 ,
```
derkling: Can we just keep the conversion "inline"?
E.g.
```python
        'pid' : int(match.group('pid')) if match.group('pid') else None
```

trappy/systrace.py
```diff
@@ -88,9 +88,12 @@ def generate_data_dict(self, data_str):
...
+                          'func' : match.group('func' ),
+                          'data' : match.group('data' ) }
```
derkling: Can we also move this closing `}` to a new line with one less indentation?

derkling:
This PR remembered me we should add also some testcase to cover some example of possible android injected events.

Can you add a patch where you extend `libs/trappy/tests/trace_systrace.html` with an example of each possible event we currently generate from user-space and correspondingly a test case which check we are parsing all the `pid`, `func` and `data` fields as expected?

trappy/systrace.py
joelagnel: Ok sure

trappy/systrace.py
joelagnel: sure

joelagnel:
Add a few more test cases and fixed up some style comments.

joelagnel:
@sinkap over to you. All changes requested are done

tests/test_systrace.py
```diff
@@ -55,14 +57,32 @@ def test_cpu_counting(self):
...
+        # Test a 'B' event (begin)
         trace = trappy.SysTrace("trace_sf.html")
         dfr = trace.tracing_mark_write.data_frame
-        self.assertTrue(dfr['__pid'].iloc[2], 7459)
-        self.assertTrue(dfr['__comm'].iloc[2], 'RenderThread')
-        self.assertTrue(dfr['pid'].iloc[2], 7459)
-        self.assertTrue(dfr['event'].iloc[2], 'B')
-        self.assertTrue(dfr['func'].iloc[2], 'notifyFramePending')
-        self.assertTrue(dfr['data'].iloc[-2], 'HW_VSYNC_0')
+        self.assertEquals(dfr['__pid'].iloc[2], 7591)
```
derkling: Instead of keep adding multiple lines for each event, perhaps something similar to what we did here:
   https://github.com/ARM-software/trappy/blob/master/tests/test_base.py#L100
can be more generic?

trappy/systrace.py
```diff
@@ -88,9 +88,14 @@ def generate_data_dict(self, data_str):
 
         match = SYSTRACE_EVENT.match(data_str)
         if match:
-            data_dict = { 'event': match.group('event'),
-                          'pid'  : match.group('pid'),
-                          'func' : match.group('func'),
-                          'data' : match.group('data') }
+            pid = match.group('pid')
```
derkling: Well, I guess these tree lines can be removed now that you added the check inline, isn't it?

joelagnel: Yes sorry, I'll get rid of them

joelagnel: Yes sorry I'll get rid of them

tests/test_systrace.py
joelagnel: Yes, probably. For now I think I will prefer to leave this as is since it involves an overhaul of this part of the test. For future test cases, will keep it in mind

joelagnel:
fixed last comment, good to go

sinkap:
Merged.

PR #253, opened 2017-06-17
Remove raw trace output

credp:
In the quest to speed up trappy parsing, I noticed that we can use the -r option for trace-cmd report so that we can use per-event raw parsing instead of generating a whole new text file.

This set of patches slightly alters the initialisation of parse objects so that we can have the set of events to use ready before we create the file, and then scans them to find which need raw formatting or not.

Then I use that to generate a single file. The patches look a little scrappy, but it works for me using the 
trappy example trace.

### DIFF
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -511,6 +511,19 @@ class FTrace(GenericFTrace):
         self.__populate_metadata()
         self._do_parse()
 
+    def __warn_about_txt_trace_files(self, trace_dat, raw_txt, formatted_txt):
+        self.__get_raw_event_list()
+        warn_text = ( "You appear to be parsing both raw and formatted "
+                      "trace files. TRAPpy now uses a unified format. "
+                      "If you have the {} file, remove the .txt files "
+                      "and try again. If not, you can manually move "
+                      "lines with the following events from {} to {} :"
+                      ).format(trace_dat, raw_txt, formatted_txt)
+        for raw_event in self.raw_events:
+            warn_text = warn_text+" \"{}\"".format(raw_event)
+
+        raise RuntimeError(warn_text)
+
     def __process_path(self, basepath):
         """Process the path and return the path to the trace text file"""
 
@@ -520,9 +533,13 @@ class FTrace(GenericFTrace):
             trace_name = os.path.join(basepath, "trace")
 
         trace_txt = trace_name + ".txt"
+        trace_raw_txt = trace_name + ".raw.txt"
         trace_dat = trace_name + ".dat"
 
         if os.path.isfile(trace_dat):
+            # Warn users if raw.txt files are present
+            if os.path.isfile(trace_raw_txt):
+                self.__warn_about_txt_trace_files(trace_dat, trace_raw_txt, trace_txt)
             # TXT traces must always be generated
             if not os.path.isfile(trace_txt):
                 self.__run_trace_cmd_report(trace_dat)
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -506,7 +506,8 @@ class FTrace(GenericFTrace):
         self.needs_raw_parsing = True
         super(FTrace, self).__init__(name, normalize_time, scope, events,
                                      window, abs_window)
-        self.trace_path, self.trace_path_raw = self.__process_path(path)
+        self.raw_events = []
+        self.trace_path = self.__process_path(path)
         self.__populate_metadata()
         self._do_parse()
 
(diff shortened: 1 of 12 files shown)


credp:
I did some updates in order to pass the tests. I found a bug in systrace parsing, and the rest of the changes are in the tests themselves as they are not testing the correct things when you have raw and formatted events mixed in together rather than in separate files.

sinkap:
From offline (er.. this is more offline :) ) discussion.

- Add an exception when the trace does not have a .dat file and hits a raw event
- Change the `doParse` to _do_parse.

This is really awesome. Good riddance `.raw.txt` !

credp:
I pushed an update with the stuff we talked about earlier. It passes all the tests for me, fingers crossed.

This is what you get if you run against an old trace.raw.txt file.

> RuntimeError: You appear to be parsing both raw and formatted trace files. TRAPpy now uses a unified format. If you have the ./example_results/trace.dat file, remove the .txt files and try again. If not, you can manually move lines with the following events from ./example_results/trace.raw.txt to ./example_results/trace.txt : "sched_wakeup_new" "sched_switch" "sched_wakeup"

trappy/ftrace.py
```diff
@@ -499,32 +500,35 @@ def __process_path(self, basepath):
...
+        for event_class in (self.thermal_classes, self.sched_classes, self.dynamic_classes):
+            for trace_class in event_class.itervalues():
+                raw = getattr(trace_class, 'parse_raw', None)
+                if raw:
+                    name = getattr(trace_class, 'name', None)
+                    if name:
+                        self.raw_events.append(name)
 
     def __run_trace_cmd_report(self, fname):
         """Run "trace-cmd report fname > fname.txt"
```
derkling: This should probably be updated to be:

```python
"""Run "trace-cmd report [-r raw_event]* fname > fname.txt
```

and we can also add a sentence in the following paragraph explaining the usage of the `-r` flag

trappy/ftrace.py
```diff
@@ -534,8 +538,13 @@ def __run_trace_cmd_report(self, fname):
         if not os.path.isfile(fname):
             raise IOError("No such file or directory: {}".format(fname))
 
-        raw_trace_output = os.path.splitext(fname)[0] + ".raw.txt"
         trace_output = os.path.splitext(fname)[0] + ".txt"
+        # Ask for the raw event list and request them unformatted
+        self.__get_raw_event_list()
+        for raw_event in self.raw_events:
+            cmd.append('-r')
```
derkling: what about:
```python
for raw_event in self.raw_events:
    cmd.extend(['-r', raw_event])
```

trappy/ftrace.py
```diff
@@ -491,6 +491,19 @@ def __init__(self, path=".", name="", normalize_time=True, scope="all",
         self.__populate_metadata()
         self._do_parse()
 
+    def __warn_about_txt_trace_files(self, trace_dat, raw_txt, formatted_txt):
+        self.__get_raw_event_list()
```
derkling: What about moving the format string at the bottom of the source where we can better format it.
Something like:

```python
def __warn_about_txt_trace_files(self, trace_dat, raw_txt, formatted_txt):
    #...
    TXT_TRACE_WARN_FMT.format(trace_dat,
                              ', '.join(raw_event),
                              raw_txt, formatted_txt)
    #...

# At end of file:
TXT_TRACE_WARN_FMT = """
You appear to be parsing both raw and formatted trace files.

TRAPpy now uses a unified format, if you have the {} file, remove ALL the .txt
files and try again.

If not, you can manually move lines with these events:
   {}
from {} to {}.
"""
```

trappy/ftrace.py
```diff
@@ -124,6 +119,13 @@ def unregister_parser(cls, cobject):
             if cobject == obj:
                 del scope_classes[name]
 
+    def _do_parse(self):
```
credp: No, then it becomes a class-private function and you can't call it from inherited classes :) I spent a little while finding that out last night!

trappy/ftrace.py
```diff
@@ -75,15 +75,10 @@ def __init__(self, name="", normalize_time=True, scope="all",
...
-        self.__parse_trace_file(self.trace_path, window, abs_window)
-        if self.needs_raw_parsing and (self.trace_path_raw is not None):
-            self.__parse_trace_file(self.trace_path_raw, window, abs_window,
-                                    raw=True)
-        self.finalize_objects()
-
-        if normalize_time:
-            self.normalize_time()
+        # save parameters to complete init later
+        self.normalize_time_param = normalize_time
```
credp: I think I can't make it completely private, as it is in a base class and called from derived classes, but I'll change it around a bit and fix up the tests which refer directly to it.

trappy/ftrace.py
```diff
@@ -499,32 +500,35 @@ def __process_path(self, basepath):
...
+            if not os.path.isfile(trace_txt):
                 self.__run_trace_cmd_report(trace_dat)
-            # TXT (and RAW) traces must match the most recent binary trace
+            # TXT traces must match the most recent binary trace
             elif os.path.getmtime(trace_txt) < os.path.getmtime(trace_dat):
                 self.__run_trace_cmd_report(trace_dat)
 
-        if not os.path.isfile(trace_raw):
-            trace_raw = None
+        return trace_txt, None
```
credp: Yep. Will sort out now. :+1:

trappy/sched.py
```diff
@@ -104,9 +104,10 @@ class SchedSwitch(Base):
     """Parse sched_switch"""
 
     unique_word = "sched_switch:"
+    parse_raw = True
```
credp: I think I will leave that for a future enhancement if you don't mind :)

trappy/ftrace.py
credp: I didn't address this yet.

trappy/ftrace.py
credp: I just realised I didn't address this yet either.

credp: Done

trappy/ftrace.py
credp: Done

credp:
OK, last 2 comments addressed. New branch pushed.

joelagnel:
LGTM. Makes the who report generation less confusing.
Chris, could you pleasre rebase on top of https://github.com/ARM-software/trappy/pull/247 since that was pushed first and if @bjackman is back with his last fix for that, then merging that is imminent :)

joelagnel:
@credp I love these patches the more I look at them. Also hopefully rebasing them ontop of the caching ones isn't much trouble for you, but let me know if I can help with that!
Also I figure at some point we should add time-based tests as well to ensure speed up related fixes don't regress!

credp:
I was looking at the issue list earlier, and I think this should fix issue #210.

trappy/ftrace.py
derkling: Right, I've missed its usage from sub-classes.

trappy/sched.py
derkling: Sure! ;-)

derkling:
@joelagnel performance regression tests would be awesome... but I doubt it's something easy to implement, especially considering the different platform we use for development.

Perhaps we can just point people to a "golden trace" and ask every contributor to cross-check using this trace that there are not sensible time regressions with-vs-without the patches of a pull request.

Do you have any other idea specifically in mind? In any case we should move this discussion into a new issue: can you do that?

derkling:
@credp I agree, this PR fixes the fundamental issue described in #210.

joelagnel:
@derkling yes you are right its hard to implement. I will create a new issue and we can chat there.

sinkap:
Merged! :+1:

Issue #250, opened 2017-06-12
Add a generic DF join API


derkling:
In many analysis it happens that we are interested in joining information coming from different DF.
For example, let say we have a trace like this:

```
            adbd-5709  [007]  2943.184105: sched_contrib_scale_f: cpu=7 cpu_scale_factor=1
            adbd-5709  [007]  2943.184105: sched_load_avg_cpu:   cpu=7 util_avg=825
     ->transport-5713  [006]  2943.184106: sched_load_avg_cpu:   cpu=6 util_avg=292
     ->transport-5713  [006]  2943.184107: sched_contrib_scale_f: cpu=6 cpu_scale_factor=2
            adbd-5709  [007]  2943.184108: sched_load_avg_cpu:   cpu=7 util_avg=850
            adbd-5709  [007]  2943.184109: sched_contrib_scale_f: cpu=7 cpu_scale_factor=3
            adbd-5709  [007]  2943.184110: sched_load_avg_cpu:   cpu=6 util_avg=315
```

Currently we can easily build two DF, one for sched_load_avg_cpu and another for sched_contrib_scale_f.

However, in some analysis it could be useful and correlate the information from these two events, thus getting a single DF where we see a consistent view of the most updated information from both.

In these cases we have a **"master_df"**, e.g. sched_load_avg_cpu, where we want to propagate into the information from a **"secondary_df"**, e.g. sched_contrib_scale_f.

This would require to:

1. Join the master_df with the secondary_df

2. Fix any index collision eventually happening, for example in the previous small trace we can see that at the exact time 2943.184105 we have one event for both master_df and secondary_df on each CPU.

A join of these two DF should grant that:
   - the order of the events is consistent with the trace ordering, i.e. sched_contrib_scale_f should be before sched_load_avg_cpu in CPU7 but after in CPU6
   - the time difference between the two events should be almost not noticiable, thus probably we should fix the overlapping timestamps by adding one 1ns to each duplicated index, thus removing the index collision without risking to create a new one with a following event.

Than we need to:

3. forward propagate each secondary_df columns by considering the value of a "pivot" column which is shared among the two DFs, for example the value `cpu` can be used to forward propagate the others `sched_contrib_scale_f` columns (i.e. `freq_scale_factor` and `cpu_scale_factor`) in the `sched_load_avg_cpu` rows

4. remove all the secondary_df rows which values have been already properly propagated in the following primary_df rows

All these operations together should be supported by a new generic convenience API which, once called with something like:

[...]

PR #254, opened 2017-06-19
Add support for trace event processing using callbacks

cobrien7:
Some trace analysis tasks are most easily expressed by iterating over
events in chronological order. Add a run_custom_analysis() method to
the GenericFTrace class that will run user-defined functions on trace
events in the order they occurred.

Signed-off-by: Connor O'Brien <connoro@google.com>

### DIFF
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -315,6 +315,55 @@ is part of the trace.
 
         return ret
 
+    def apply_callbacks(self, fn_map):
+        """
+        Apply callback functions to trace events in chronological order.
+
+        This method iterates over a user-specified subset of the available trace
+        event dataframes, calling different user-specified functions for each
+        event type. These functions are passed a dictionary mapping 'Index' and
+        the column names to their values for that row.
+
+        For example, to iterate over trace t, applying your functions callback_fn1
+        and callback_fn2 to each sched_switch and sched_wakeup event respectively:
+
+        t.apply_callbacks({
+            "sched_switch": callback_fn1,
+            "sched_wakeup": callback_fn2
+        })
+        """
+        dfs = {event: getattr(self, event).data_frame for event in fn_map.keys()}
+        events = [event for event in fn_map.keys() if not dfs[event].empty]
+        iters = {event: dfs[event].itertuples() for event in events}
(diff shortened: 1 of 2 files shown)


cobrien7:
This patch should work exactly the same as https://github.com/ARM-software/lisa/pull/418, except that the new method is added to GenericFTrace instead of the LISA Trace class.

joelagnel:
Hi Connor, Is it possible to not use the word 'analysis' in the core function? If I understand correctly, trappy is used for data generation and visualization, analysis usage of this will be in LISA. Also this feature may be used for https://github.com/ARM-software/trappy/issues/250. Considering this, I would call it 'run_event_callbacks'.

joelagnel:
CC @cobrien7

derkling:
I would say that `TRAPpy` is mainly for data generation... while analysis AND visualisation is mainly in LISA. We know have some plotting methods in TRAPpy which are still there mainly for backward compatibility reasons., but future bits should be better organised according to a schema where everything which is not "generic DF generation and processing" should *not* be in TRAPpy.

derkling:
Regarding the function name, what about something like a quite generic `process_events`?

joelagnel:
Hmm @derkling I prefer a more explicit name like `run_event_callbacks`. @cobrien7 could you update this PR with the change?

joelagnel:
Never mind, so @cobrien7 did change it. @sinkap I guess over to you then.

trappy/ftrace.py
```diff
@@ -318,6 +318,55 @@ def get_all_freqs_data(self, map_label):
 
         return ret
 
+    def run_event_callbacks(self, fn_map):
```
derkling: Again, I'm not really convinced about this name... maybe because of its linkage with the old "compare and run" naming used for the original TRAPpy code... @sinkap up to you to proposa a more generic name for this functionality ;-)

sinkap: I guess the answer lies in the callback. How about `apply_callbacks`?

joelagnel:
`apply_callbacks` sounds good to me.

sinkap:
Done! :+1:

PR #255, opened 2017-06-20
trappy/ftrace: add support for boot clock timestamp

derkling:
FTrace can be configured to report events timestamp using different
clock sources which can be selected via the trace_clock sysfs attribute
as described in:
  https://www.kernel.org/doc/Documentation/trace/ftrace.txt

The global clock source reports time in [s] with a [us] resolution.
Other sources instead, like for example the boot clock, uses [ns].
Thus, in these last cases we do not have decimals in the timestamp.

Let's update the special fields regexp to match both [s] and [ns]
formatted times. This also update the base test to add a set of trace
events which are expressed in [ns] resolution.

### DIFF
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -49,7 +49,7 @@ def _plot_freq_hists(allfreqs, what, axis, title):
 SPECIAL_FIELDS_RE = re.compile(
                         r"^\s*(?P<comm>.*)-(?P<pid>\d+)(?:\s+\(.*\))"\
                         r"?\s+\[(?P<cpu>\d+)\](?:\s+....)?\s+"\
-                        r"(?P<timestamp>[0-9]+\.[0-9]+): (\w+:\s+)+(?P<data>.+)"
+                        r"(?P<timestamp>[0-9]+(?P<us>\.[0-9]+)?): (\w+:\s+)+(?P<data>.+)"
 )
 
 class GenericFTrace(BareTrace):
@@ -199,7 +199,14 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
             comm = fields_match.group('comm')
             pid = int(fields_match.group('pid'))
             cpu = int(fields_match.group('cpu'))
+
+            # The timestamp, depending on the trace_clock configuration, can be
+            # reported either in [s].[us] or [ns] format. Let's ensure that we
+            # always generate DF which have the index expressed in:
+            #    [s].[decimals]
             timestamp = float(fields_match.group('timestamp'))
+            if not fields_match.group('us'):
+                timestamp /= 1e9
             data_str = fields_match.group('data')
 
             if not self.basetime:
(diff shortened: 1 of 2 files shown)


derkling:
Maybe @joelagnel can be interested in having a look since I think he had a similar "issue" with some Android traces in the past.

bjackman:
The documentation says "the timestamp in <secs>.<usecs> format" so I wonder if this is a bug in a device kernel? If it's not going to be fixed in the kernel then yeah I guess we have to support it here.

trappy/ftrace.py
```diff
@@ -49,7 +49,7 @@ def _plot_freq_hists(allfreqs, what, axis, title):
 SPECIAL_FIELDS_RE = re.compile(
                         r"^\s*(?P<comm>.*)-(?P<pid>\d+)(?:\s+\(.*\))"\
                         r"?\s+\[(?P<cpu>\d+)\](?:\s+....)?\s+"\
-                        r"(?P<timestamp>[0-9]+\.[0-9]+): (\w+:\s+)+(?P<data>.+)"
+                        r"(?P<timestamp>[0-9]+(\.[0-9]+)?): (\w+:\s+)+(?P<data>.+)"
```
bjackman: We need to detect if the timestamps are in the ns format and then convert the timestamps to seconds, otherwise the data will look all wrong.

joelagnel: I was going to say same thing

derkling:
@bjackman the only reference I can find related to the "." format is related to the specific description of the function tracer, I'm looking at this file:
   https://www.kernel.org/doc/Documentation/trace/ftrace.txt
... thus it could be very well a post-processing formatting.

Moreover, the description of the `trace_clock` parameter says that the `boot` clock comes from
`ktime_get_boot_fast_ns()`... which name suggests it should return a ns based timestamp. There is also an explicit reference to possible required post-processing.

derkling:
Update:
- detect [s].[us] format using an additional regexp capture group
- pre-process [ns] timestamps to convert them in [s].[decimals] format
- update test_base to properly check for timestamps which are always expressed in the [s].[decimals] foramt

credp:
As I understand it the timestamp format in a trace depends upon the configured trace_clock - it's perfectly possible and valid to use the nanosecond boot clock.

trappy/ftrace.py
```diff
@@ -200,6 +200,8 @@ def contains_unique_word(line, unique_words=cls_for_unique_word.keys()):
             pid = int(fields_match.group('pid'))
             cpu = int(fields_match.group('cpu'))
             timestamp = float(fields_match.group('timestamp'))
+            if not fields_match.group('us'):
```
bjackman: Could do with a comment here explaining that the timestamp is either [s].[us] or [ns].

derkling:
Update:
- added a comment to explain the timestamp adjustments

joelagnel:
LGTM... Let's get it merged ASAP 😁😁

sinkap:
Merged.

PR #256, opened 2017-06-25
trappy: 30% speed up goodies !!!

joelagnel:
Some speed enhancements, summarizing, altogether giving 30-35% improvement in performance:
* Optimize unique word search.
* Optimize search for array pattern in data.
* Optimize integer conversion in generate_data_dict.

### DIFF
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -179,6 +179,27 @@ class Base(object):
         self.line_array.append(line)
         self.data_array.append(data)
 
+    def string_cast(self, string, type):
+        """ Attempt to convert string to another type
+
+        Here we attempt to cast string to a type. Currently only
+        integer conversion is supported with future expansion
+        left open to other types.
+
+        :param string: The value to convert.
+        :type string: str
+
+        :param type: The type to convert to.
+        :type type: type
+        """
+        # Currently this function only supports int conversion
+        if type != int:
+            return
+        # Handle false-positives for negative numbers
+        if not string.lstrip("-").isdigit():
+            return string
+        return int(string)
+
     def generate_data_dict(self, data_str):
         data_dict = {}
         prev_key = None
@@ -190,10 +211,7 @@ class Base(object):
                 data_dict[prev_key] += ' ' + field
                 continue
             (key, value) = field.split('=', 1)
-            try:
-                value = int(value)
-            except ValueError:
-                pass
+            value = self.string_cast(value, int)
             data_dict[key] = value
             prev_key = key
         return data_dict
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -165,6 +165,15 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
                 trace_class = DynamicTypeFactory(event_name, (Base,), kwords)
                 self.class_definitions[event_name] = trace_class
 
+    def __get_trace_class(self, line, cls_word):
+        trace_class = None
+        for unique_word, cls in cls_word.iteritems():
+            if unique_word in line:
+                trace_class = cls
+                if not cls.fallback:
+                    return trace_class
+        return trace_class
+
     def __populate_data(self, fin, cls_for_unique_word):
         """Append to trace data from a txt trace"""
 
@@ -173,16 +182,10 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
                                            actual_trace)
 
         for line in actual_trace:
-            trace_class = None
-            for unique_word, cls in cls_for_unique_word.iteritems():
-                if unique_word in line:
-                    trace_class = cls
-                    if not cls.fallback:
-                        break
-            else:
-                if not trace_class:
-                    self.lines += 1
-                    continue
+            trace_class = self.__get_trace_class(line, cls_for_unique_word)
+            if not trace_class:
+                self.lines += 1
+                continue
 
             line = line[:-1]
 
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -215,7 +215,8 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
                 return
 
             # Remove empty arrays from the trace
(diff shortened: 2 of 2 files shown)


trappy/ftrace.py
```diff
@@ -168,28 +168,21 @@ class definitions list.  Otherwise, register a class to parse
...
-
         actual_trace = itertools.dropwhile(self.trace_hasnt_started(), fin)
         actual_trace = itertools.takewhile(self.trace_hasnt_finished(),
                                            actual_trace)
 
         for line in actual_trace:
-            if not contains_unique_word(line):
-                self.lines += 1
-                continue
+            trace_class = None
```
derkling: While we are at that, maybe we can cleanup this further by moving the following block into a local function, which replaces the one you remove above, e.g.

```python
def get_trace_class(line):
    trace_class = None
    for unique_word, cls in cls_for_unique_word.iteritems():
        if unique_word in line:
            trace_class = cls
            if not cls.fallback:
                return trace_class
    return trace_class
```

With that, the beginning of the parsing loop should be more simple "to parse":

```python
for line in actual_trace:
    trace_class = self.get_trace_class(line)
    if not trace_class:
        # Line not matching any of the enabled/registered attention words
        # still we account for the skipped trace line
        self.line += 1
        continue
    line = line[:-1]
   #...
```

trappy/ftrace.py
```diff
@@ -215,7 +215,8 @@ def __populate_data(self, fin, cls_for_unique_word, window, abs_window):
                 return
 
             # Remove empty arrays from the trace
-            data_str = re.sub(r"[A-Za-z0-9_]+=\{\} ", r"", data_str)
+            if "={}" in data_str:
```
derkling: Super cool :heart_eyes: 

I would just suggest to strip down the commit title to something shorter, e.g.
```
trappy: skip empty array regex for common case
```
kudos for the big speedup are still reported in the change-log ;-)

trappy/base.py
```diff
@@ -179,6 +179,12 @@ def append_data(self, time, comm, pid, cpu, line, data):
         self.line_array.append(line)
         self.data_array.append(data)
 
+    def conv_to_int(self, value):
```
derkling: Maybe we can use a less ugly name as well as make this function a bit more generic just by changing its name to something like `type_cast(self, value)`.

Than it would be easier for a future patch to add in additional type casting (e.g. floats, hex)... potentially we can also map-translate known string tokens to values (e.g. for sched_switch) status events...

joelagnel:
CC @sinkap

trappy/ftrace.py
joelagnel: Sure, thanks ;-)
Yes I'll shorten the commit title.

trappy/ftrace.py
joelagnel: Ok sounds good, I'll just add the cleanup PR ontop of the original one to keep it less confusing.

trappy/base.py
joelagnel: I think we can make the handling of other types as a future addition if you don't mind. Also about the naming, I will change it to `cast_to_int()`. Hope that sounds less ugly.

derkling:
I'm happy for this to be merged... ASAP ;-)

trappy/base.py
sinkap: What we are really doing is parsing a string to an integer. I would name it parse_int(str)

trappy/base.py
```diff
@@ -179,6 +179,12 @@ def append_data(self, time, comm, pid, cpu, line, data):
...
+        # Handle false-positives for negative numbers
+        if value.lstrip("-").isdigit():
+            value = int(value)
```
sinkap: I would still check and tolerate a `ValueError` here.

```
try:
  value = int(value)
except:
  pass

There can be other cases where this might fail.

File: trappy/ftrace.py [hunk 5]
```diff
@@ -168,28 +168,21 @@ class definitions list.  Otherwise, register a class to parse
...
             for unique_word, cls in cls_for_unique_word.iteritems():
                 if unique_word in line:
                     trace_class = cls
                     if not cls.fallback:
                         break
             else:
                 if not trace_class:
-                    raise FTraceParseError("No unique word in '{}'".format(line))
+                    self.lines += 1
+                    continue
```
sinkap: This seems to be again changing the behaviour of the code. We have a unique word but no class for that.

derkling: The assumption here is that we loop on all the events we know about and we are able to parse, if none matches their unique word that means we can skip this line while still counting for the trace line we parsed.

That's why we replaced the exception with a continue.

Moreover, I think the previous `raise` was already dead-code. Indeed, we was checking for a `unique_word` match using `contains_unique_word` right before the loop, thus if we enter the loop we will certainly set `trace_class` to some (hopefully not null) `cls`.

trappy/base.py
bjackman: Sorry, can't resist bikeshedding here... shouldn't it be "maybe_parse_int" since we're only parsing an integer if it's an integer?

joelagnel: Ok. How about we name it `try_parse_int`? I think that should make everyone happy hopefully

trappy/base.py
joelagnel: Can you give an example? I would rather it failed if `isdigit` returned true and it still failed because then we can find those cases and make it more robust. Also *I think* using try/except will slow it down so I'd rather not do it in the hot path...

derkling:
@bjackman remark makes me suggesting one more time to `s/conv_to_int/type_cast/`... which is much more generic as well as ready for future extension.

joelagnel:
Ok. I reread your earlier comment and I agree with you now. Good idea. So, @sinkap @bjackman, I will call it `type_cast` which we can expand in future to detect other types. We are indeed changing the type from string to possibly something else..

joelagnel:
So I call it `string_cast` type `type` as a parameter which makes it generic and handling only `int` conversion for now. Also fixed the other comments. I think this should be good to go. @sinkap

sinkap:
And it's a goal! :+1:

PR #258, opened 2017-06-29
trappy: add support to parse TGID in systrace

joelagnel:
TGID is a fundamental property tracked in systrace for each trace record. Add
support to parse it if its available.

Change-Id: Ie79698d90e0406cc11c52d364144ec08c33dfac4


joelagnel:
CC @sinkap

bjackman:
Just a note, I think the failure is because of a Python bug: https://stackoverflow.com/questions/3675144/regex-error-nothing-to-repeat

It might be fixed by replacing `\s` with `[ \t]`?

PR #260, opened 2017-07-08
trappy/ftrace: merge primary and secondary DFs based on pivot

joelagnel:
I hacked something up that's working, just posting here. Basic idea is concatenate the DFS, mark them as primary/secondary, walk through them and remember the secondary information, propagate it to primary and rebuild the primary DF. (CC @derkling @bjackman @sinkap)


trappy/utils.py
```diff
@@ -102,3 +105,55 @@ def handle_duplicate_index(data,
             dup_index_left += 1
 
     return data.reindex(new_index)
+
+def merge_dfs(pr_df, sec_df, pivot):
```
sinkap: Please add a doctsring for this method and explain with an example in the documentation. 

(A short excerpt like the one you have in the gist.)

trappy/utils.py
```diff
@@ -102,3 +105,55 @@ def handle_duplicate_index(data,
...
+            return
+
+        # Propogate latest secondary info
+        for key, value in data.iteritems():
+            if key == pivot:
+                continue
+            try:
+                if np.isnan(value):
+                    data[key] = pivot_map[data[pivot]][key]
+            except:
```
sinkap: A general except is not good. Can you specifically catch the excpetions that can possibly occur (and you expcect them to occur)

trappy/utils.py
sinkap: Also, since this is core functionality. We would need tests as well.

joelagnel: Ok will do

joelagnel: Yep this is a bit of a preview so I didn't add yet. Will do, thanks

trappy/utils.py
joelagnel: I got rid of this exception altogether to fix a perf issue and will post the fix soon, thanks

PR #261, opened 2017-07-09
cache: Fix issue where single missing CSV deletes all

joelagnel:
There's a bug in trappy caching where corruption or removal of single CSVs
causes all CSVs to get removed. Fix this and add a test-case to cover it.
Also test that only missing items are regenerated.

Change-Id: Iea41c8e0a005515790b580e7f78f06bdd60141a3


sinkap:
:+1: For the test case! Thanks! Merged.

PR #263, opened 2017-08-07
bare_trace: fix signature of generate_data_dict

mdigiorgio:
Classes that implement the generate_data_dict method have an additional argument
to the signature. This leads to errors in case the base class method is called
with the additional argument.

### DIFF
--- a/tests/test_ftrace.py
+++ b/tests/test_ftrace.py
@@ -351,6 +351,26 @@ class TestFTrace(BaseTestThermal):
         self.assertEquals(trace.thermal.data_frame.iloc[0]["temp"], 68989)
         self.assertEquals(trace.thermal.data_frame.iloc[-1]["temp"], 69530)
 
+    def test_parse_tracing_mark_write_events(self):
+        """Check that tracing_mark_write events are parsed without errors"""
+
+        in_data = """     sh-1379  [002]   353.397813: print:                tracing_mark_write: TRACE_MARKER_START
+     shutils-1381  [001]   353.680439: print:                tracing_mark_write: cpu_frequency:        state=450000 cpu_id=5"""
+
+        with open("trace.txt", "w") as fout:
+            fout.write(in_data)
+
+        try:
+            trace = trappy.FTrace()
+        except TypeError as e:
+            self.fail("tracing_mark_write parsing failed with {} exception"\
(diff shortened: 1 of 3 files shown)


tests/test_ftrace.py
```diff
@@ -351,6 +351,21 @@ def test_ftrace_accepts_abs_window(self):
...
+
+        with open("trace.txt", "w") as fout:
+            fout.write(in_data)
+
+        try:
+            trace = trappy.FTrace()
+        except TypeError as e:
+            self.fail("tracing_mark_write parsing failed with {} exception"\
+                      .format(e.message))
+
```
derkling: Should we not add also something like:

```python
self.assertEquals(trace.tracing_mark_write.data_frame.iloc[-1]["string"], "TRACE_MARKER_START")
self.assertEquals(len(trace.tracing_mark_write.data_frame), 2)
```

mdigiorgio: Sounds like a good idea, I've just added it.

tests/test_ftrace.py
```diff
@@ -351,6 +351,24 @@ def test_ftrace_accepts_abs_window(self):
...
+        with open("trace.txt", "w") as fout:
+            fout.write(in_data)
+
+        try:
+            trace = trappy.FTrace()
+        except TypeError as e:
+            self.fail("tracing_mark_write parsing failed with {} exception"\
+                      .format(e.message))
+        self.assertEquals(trace.tracing_mark_write.data_frame.iloc[-1]["string"],
```
bjackman: I might be being silly here.. shouldn't `iloc[-1]` be the cpu_frequency event?

bjackman: Oh, you might also have a problem because of the 'cpu_frequency' parser gobbling the second event (in which case I think `iloc[-1]` would be correct but the `len` would be 1

mdigiorgio: Right, the second case is correct. Pushed new version, thanks.

bjackman: Sorry could you just add a comment explaining where the cpu_frequency event is gone? Might not be obvious.

bjackman:
Can we merge this?

credp:
Lgtm

PR #264, opened 2017-08-17
ftrace: Disable caching by default

bjackman:
There are some issues with caching at the moment which I haven't been
able to debug but that are confirmed as affecting multiple
people. Until we can find the time to fix the issue, let's disable
caching by default.


joelagnel:
Ack.

PR #265, opened 2017-08-17
base: Improve errors for failures to parse

bjackman:
Some events from the kernel cannot be parsed by old trace-cmd
binaries. This results in data_str starting with "[FAILED TO PARSE]",
which we ourselves fail to parse because we expect to find a
key=value pair.

This adds a missing 'continue' statement to avoid attempting to avoid
a KeyError when attempting to use an unset prev_key as a key into
data_dict, and also adds a special error for the "[FAILED TO PARSE]"
case.

### DIFF
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -21,6 +21,8 @@ import warnings
 
 from resource import getrusage, RUSAGE_SELF
 
+from trappy.exception import TrappyParseError
+
 def _get_free_memory_kb():
     try:
         with open("/proc/meminfo") as f:
@@ -206,6 +208,18 @@ class Base(object):
         prev_key = None
         for field in data_str.split():
             if "=" not in field:
+                if not prev_key:
+                    if 'FAILED TO PARSE' in data_str:
+                        warnings.warn(
+                            'trace-cmd failed to parse the "{}" event. You may '
+                            'need to compile the latest trace-cmd and put it in '
+                            'your $PATH. Continuing, but some data may be missing'
+                            .format(self.unique_word))
+                        continue
+                    else:
+                        raise TrappyParseError(
+                            "TRAPpy's parser for '{}' failed to parse the line:"
+                            "\n{}".format(self.unique_word, data_str))
                 # Concatenation is supported only for "string" values
                 if type(data_dict[prev_key]) is not str:
                     continue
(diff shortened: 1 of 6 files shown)


trappy/base.py
```diff
@@ -206,6 +206,16 @@ def generate_data_dict(self, data_str):
         prev_key = None
         for field in data_str.split():
             if "=" not in field:
+                if not prev_key:
+                    if 'FAILED TO PARSE' in data_str:
+                        raise ValueError(
+                            'trace-cmd failed to parse the "{}" event. You may '
+                            'need to compile the latest trace-cmd and put it in '
+                            'your $PATH'.format(self.__class__.__name__))
```
valschneider: I think `self.__class__.unique_word` would be more accurate as it will give the exact event name. Class name is already pretty good though, it'll be roughly the same thing but in CamelCase.

trappy/base.py
```diff
@@ -206,6 +206,16 @@ def generate_data_dict(self, data_str):
...
+                    else:
+                        raise ValueError(
+                            "TRAPpy's parser for '{}' failed to parse the line:"
+                            "\n{}".format(self.__clas__.__name__, data_str))
```
valschneider: Ditto for event name

bjackman:
Phew.. I really really really don't want to break anything here so trying to be a bit hyper-cautious with testing.. added a unit test, I also added a separate exception class just because there was already one instance of that in the code.

bjackman:
OK, changed the FAILED TO PARSE case to just print a warning and continue. There might be missing data, but at least we know what went wrong and that the rest of the trace is probably fine.

trappy/base.py
```diff
@@ -206,6 +208,18 @@ def generate_data_dict(self, data_str):
         prev_key = None
         for field in data_str.split():
             if "=" not in field:
+                if not prev_key:
+                    if 'FAILED TO PARSE' in data_str:
+                        warnings.warn(
+                            'trace-cmd failed to parse the "{}" event. You may '
```
valschneider: Ideally we'd want to have each warning only once (one per failed to parse event). Can we just roll with the "warnings only printed once" thing or should we do something a bit better ?

bjackman: I think it only prints it once for each event already

valschneider: LGTM then

PR #268, opened 2017-09-18
bare_trace: Fix get_duration() for window use

valschneider:
The previous implementation only looked at the maximum timestamp
of the traces. However, if trace windows are used to analyse a
subset of the full trace, get_duration() would return an
erroneous value:

(times are normalized)
trace = <trace built with window=[5, 10]>
print trace.get_duration()
> 10.0

This commit adds a lower-bound lookup to fix this issue.

### DIFF
--- a/tests/test_baretrace.py
+++ b/tests/test_baretrace.py
@@ -56,7 +56,7 @@ class TestBareTrace(unittest.TestCase):
         trace.add_parsed_event("pmu_counter", self.dfr[0])
         trace.add_parsed_event("load_event", self.dfr[1])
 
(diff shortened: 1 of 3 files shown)


valschneider:
@bjackman Woops, I relied on the tests I wrote for LISA's internal trace class... I'll add one here.

valschneider:
@bjackman I added window tests for ftrace and also "fixed" a test for baretrace. Even without windows, the problem was visible from that test: we were checking that the returned duration was simply the latest timestamp - which was wrong.

bjackman:
Assuming the 2nd commit is needed to make the 1st commit pass, can you squash those two?

valschneider:
@bjackman Done

PR #269, opened 2017-09-20
Thermal plotting additions

valschneider:
This PR is about:
- Changing the drawstyle of plotting methods to 'steps-post'
- Adding cdev_state plotting methods for devfreq and cpufreq cooling devices.
- Adding xlim/ylim parameters to new and existing plotting methods (mostly to help in my next LISA PR)

This is what the plots look like: https://gist.github.com/valschneider/81a247b6acb69cccfe6e0ad7682914a8#file-thermal_plotting-ipynb

### DIFF
--- a/tests/test_thermal.py
+++ b/tests/test_thermal.py
@@ -139,6 +139,23 @@ class TestThermalGovernor(BaseTestThermal):
         trappy.FTrace().thermal_governor.plot_inout_power(title="Antutu")
         matplotlib.pyplot.close('all')
 
+class TestCpuOutPower(BaseTestThermal):
+    def test_get_dataframe(self):
+        dfr = trappy.FTrace().cpu_out_power.data_frame
+
+        self.assertTrue(len(dfr) > 0)
+        self.assertTrue("cdev_state" in dfr.columns)
+
+    def test_plot_cdev_states(self):
+        """Test CpuOutPower.plot_cdev_states()
+
+        Can't check that the graph is ok, so just see that the method doesn't blow up"""
+        out = trappy.FTrace().cpu_out_power
+
+        out.plot_cdev_states()
+        out.plot_cdev_states(xlim=(3, 6))
+        matplotlib.pyplot.close('all')
+
 class TestEmptyThermalGovernor(unittest.TestCase):
     def setUp(self):
         self.previous_dir = os.getcwd()
--- a/tests/test_thermal.py
+++ b/tests/test_thermal.py
@@ -156,6 +156,23 @@ class TestCpuOutPower(BaseTestThermal):
         out.plot_cdev_states(xlim=(3, 6))
         matplotlib.pyplot.close('all')
 
+class TestDevfreqOutPower(BaseTestThermal):
+    def test_get_dataframe(self):
+        dfr = trappy.FTrace().devfreq_out_power.data_frame
+
+        self.assertTrue(len(dfr) > 0)
+        self.assertTrue("cdev_state" in dfr.columns)
+
+    def test_plot_cdev_states(self):
+        """Test CpuOutPower.plot_cdev_states()
+
+        Can't check that the graph is ok, so just see that the method doesn't blow up"""
+        out = trappy.FTrace().devfreq_out_power
+
+        out.plot_cdev_states()
+        out.plot_cdev_states(xlim=(3, 6))
+        matplotlib.pyplot.close('all')
+
 class TestEmptyThermalGovernor(unittest.TestCase):
     def setUp(self):
         self.previous_dir = os.getcwd()
--- a/trappy/cpu_power.py
+++ b/trappy/cpu_power.py
@@ -93,6 +93,33 @@ def num_cpus_in_mask(mask):
 
     return bin(value).count("1")
 
+def sanitize_mask(mask):
+    """Return mask string to usable int value """
+    return int(mask.replace(",", ""), 16)
+
+def cpus_to_mask(cpus):
+    """Returns the integer value of the list -> mask translation"""
+    res = 0
+    for cpu in cpus:
+        res |= (1 << cpu)
+    return res
+
+def mask_to_cpus(mask):
+    """Returns the list of the cpus present in the mask"""
+    res = []
+    # Remove '0b' prefix
+    bits = list(bin(mask))[2::]
+    # Reverse to LSB first order
+    bits.reverse()
+    for idx, bit in enumerate(bits):
+        if bit == '1':
+            res.append(idx)
+    return res
+
+def stringify_mask(mask):
+    """Returns a user-friendly print of a cpumask"""
+    return "CPUs {}".format(mask_to_cpus(mask))
+
 class CpuOutPower(Base):
     """Process the cpufreq cooling power actor data in a ftrace dump"""
 
@@ -119,6 +146,64 @@ class CpuOutPower(Base):
         dfr = self.data_frame
 
         return pivot_with_labels(dfr, "freq", "cpus", mapping_label) / 1000
+
+    def plot_cdev_states(self, width=None, height=None, xlim="default",
+                         ylim="range", drawstyle="default", ax=None,
+                         title="", cpus=None):
+        """Plot the cooling device state evolution
+
+        :param width: The width of the plot
+        :type width: int
+
+        :param height: The height of the plot
+        :type height: int
+
+        :param xlim: The xlim setting of the plot.
+            See :func:`~trappy.plot_utils.set_lim`
+        :type xlim: str or tuple of int
+
+        :param ylim: The ylim setting of the plot
+            See :func:`~trappy.plot_utils.set_lim`
+        :type ylim: str or tuple of int
+
+        :param drawstyle: The drawstyle setting of the plot
+        :type drawstyle: str
+
+        :param cpus: List of cpus to plot
+            All are plotted by default
+        :type cpus: list of int
+        """
+
+        from trappy.plot_utils import plot_generic
+
+        thermal_dfr = self.data_frame.copy()
+
+        # CPUs are given as a list for convenience,
+        # but they're in string cpumask format in the trace.
+        # Thus, change format to middle-ground int value
+        thermal_dfr["cpus"] = thermal_dfr["cpus"].apply(sanitize_mask)
+
+        cpumasks = thermal_dfr["cpus"].unique().tolist()
+
+        # Sanitize cpus
+        if cpus is not None:
+            global_mask = cpus_to_mask(cpus)
+
+            # Find masks that match the requested CPUs
+            # This can include other CPUs
+            selected_masks = [m for m in cpumasks if m & global_mask]
+
+            thermal_dfr = thermal_dfr[
+                thermal_dfr["cpus"].isin(selected_masks)
+            ]
+
+            if len(selected_masks) == 0:
+                raise ValueError("No {} trace for CPUs {}".format(self.unique_word, cpus))
+
+        plot_generic(thermal_dfr, "cpus", fields=["cdev_state"],
+                     prettify_name=stringify_mask, width=width, height=height,
+                     xlim=xlim, ylim=ylim, drawstyle=drawstyle, ax=ax, title=title
+        )
 
 register_ftrace_parser(CpuOutPower, "thermal")
 
--- a/trappy/devfreq_power.py
+++ b/trappy/devfreq_power.py
@@ -22,7 +22,6 @@ import pandas as pd
 from trappy.base import Base
 from trappy.dynamic import register_ftrace_parser
 
-
 class DevfreqInPower(Base):
     """Process de devfreq cooling device data regarding get_power in an
 FTrace dump"""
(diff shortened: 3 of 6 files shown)


trappy/thermal.py
```diff
@@ -88,7 +88,7 @@ def plot_temperature(self, control_temperature=None, title="", width=None,
             setup_plot = True
 
         temp_label = normalize_title("Temperature", legend_label)
-        (thermal_dfr["temp"] / 1000).plot(ax=ax, label=temp_label)
+        (thermal_dfr["temp"] / 1000).plot(ax=ax, label=temp_label, drawstyle="steps-post")
```
bjackman: Not sure about this, if you really want to see a plot of the temperature, then I think linear interpolation makes perfect sense. If you're looking at thermal_temperature to find out what input a kernel component was getting then I suppose steps-post makes sense, but I'd probably be happy if I had to set that manually when requesting the plot - can you make it a param the caller can set?

trappy/thermal.py
```diff
@@ -37,8 +37,8 @@ class Thermal(Base):
     """The Pivot along which the data is orthogonal"""
 
     def plot_temperature(self, control_temperature=None, title="", width=None,
-                         height=None, ylim="range", ax=None, legend_label="",
-                         tz_id=None):
+                         height=None, xlim="default", ylim="range", ax=None,
+                         legend_label="", tz_id=None):
```
bjackman: Need to update the docstring

trappy/cpu_power.py
```diff
@@ -93,6 +93,29 @@ def num_cpus_in_mask(mask):
...
+    """Returns the list of the cpus present in the mask"""
+    res = []
+    # Remove '0b' prefix
+    bits = list(bin(mask))[2::]
+    # Reverse to LSB first order
+    bits.reverse()
+    for idx, bit in enumerate(bits):
+        if bit == '1':
+            res.append(idx)
+    return res
```
bjackman: It's not important, but converting to a string does seem a bit laborious.. what about something like this?

```
return [i for i in range(number_of_bits) if mask & (1 << i)]
```
Not sure how you'd get `number_of_bits`, mabye `int(log(mask, 2)) + 1` or something. 

But yeah this doesn't actually matter I'm just being a nerd.

trappy/cpu_power.py
```diff
@@ -120,6 +143,89 @@ def get_all_freqs(self, mapping_label):
...
+
+        :param width: The width of the plot
+        :type width: int
+
+        :param height: The height of the plot
+        :type height: int
+
+        :param cpus: List of cpus to plot
+            All are plotted by default
+        :type tz_id: list of int
```
bjackman: This dosctring is out of sync with the actual interface

trappy/cpu_power.py
```diff
@@ -120,6 +143,89 @@ def get_all_freqs(self, mapping_label):
...
+        """
+        from matplotlib import pyplot as plt
+        from trappy.plot_utils import normalize_title, pre_plot_setup, post_plot_setup
```
bjackman: Um..  is this done elsewhere (importing inside a function)?

trappy/cpu_power.py
```diff
@@ -120,6 +143,89 @@ def get_all_freqs(self, mapping_label):
...
+        # Sanitize cpus
+        if cpus is not None:
+            global_mask = cpus_to_mask(cpus)
+
+            # Find masks that match the requested CPUs
+            # This can include other CPUs
+            selected_masks = []
+            for mask in available_masks:
+                if global_mask & mask:
+                    selected_masks.append(mask)
```
bjackman: You can shorten this with a list comp too: `select_masks = [m for m in available masks if m & global_mask]`

trappy/devfreq_power.py
```diff
@@ -72,4 +72,76 @@ def get_all_freqs(self):
...
+
+        :param width: The width of the plot
+        :type width: int
+
+        :param height: The height of the plot
+        :type height: int
+
+        :param devices: List of devfreq devices to plot
+            All are plotted by default
+        :type tz_id: list of str
```
bjackman: ditto

trappy/devfreq_power.py
```diff
@@ -72,4 +72,76 @@ def get_all_freqs(self):
...
+        """
+        from matplotlib import pyplot as plt
+        from trappy.plot_utils import normalize_title, pre_plot_setup, post_plot_setup
```
bjackman: ditto

trappy/devfreq_power.py
```diff
@@ -72,4 +72,76 @@ def get_all_freqs(self):
...
+
+            _ax.legend([cdev_label])
+            # Add vertical guidelines
+            _ax.yaxis.grid(True)
+            # Ensure tick interval is 1 and not something with weird decimals
+            states = df.unique().tolist()
+            _ax.set_yticks(range(min(states), max(states) + 1))
+
+            post_plot_setup(_ax, xlim=xlim, ylim=ylim)
+
```
bjackman: Would be nice to avoid the duplicated code. If it's a big effort to do that then I think we can live with it, though.

trappy/cpu_power.py
valschneider: Yeah I started off by copy/pasting the code for thermal:Thermal:plot_temperature which does this.

trappy/thermal.py
valschneider: For some reason the `ylim` wasn't referenced in the docstring, so I didn't reference the `xlim` parameter either. I suppose it'd be better if I referenced both ?

bjackman: Yeah might as well fix it while you'r'e touching this code

trappy/cpu_power.py
bjackman: OK with me then :man_shrugging:

trappy/cpu_power.py
valschneider: I realized devlib has some tooling for this:
```
def mask_to_list(mask):
    size = len(bin(mask)) - 2  # because of "0b"
    return [size - i - 1 for i in xrange(size)
            if mask & (1 << size - i - 1)]
```
This also uses a string conversion though. I also thought of doing it "C-style" but Python has made my laziness even worse...

valschneider:
@bjackman 
- Fixed pydoc typos
- Added missing pydocs
- Moved drawstyle to method parameters
- Added tests

Haven't had a look at factorizing the {cpu/dev}freq cdev state plotting code. I'll do that now.

valschneider:
Okay so I *think* the best way to do this would be to merge the devfreq/cpu power classes, so that they may share logic. Their dataframes are quite similar, but:
- I don't know how much I should assume about those similarities
- That merging is gonna take some more time

bjackman:
OK, if it's a big job then don't worry about it, the duplication isn't too bad.

trappy/cpu_power.py
```diff
@@ -120,6 +143,98 @@ def get_all_freqs(self, mapping_label):
...
+                thermal_dfr["cpus"].isin(selected_masks)
+            ]
+
+        cdevs = {}
+        # At this point we've removed the cpumasks we don't care about
+        for mask in thermal_dfr["cpus"].unique().tolist():
+            cdevs[mask] = thermal_dfr[thermal_dfr["cpus"] == mask]["cdev_state"]
+
+        # Prepare one subplot per cdev
+        ax = pre_plot_setup(width, height, nrows = len(cdevs))
```
valschneider: There's one thing I'm still a bit unsure about - the management of the axes. The other plotting methods I used as a starting point accept an `ax` parameter, and use that handle to plot stuff if it's `not None`. Thing is, I feel obligated to done one plot per cooling device because it gets unreadable otherwise (have look at the bottom of https://gist.github.com/valschneider/9f93ada029c8662ea954178906c4928f to see what I mean).

So that means that this `ax` parameter should be an array of axes if there's more than one cdev, so the user is expected to look into the dataframe data - and same goes for legends and titles, which is left to the user in the pre-existing methods. I went with a more "ignorant user" approach, but it may clash with the logic that has been followed for the other plots.

bjackman: Hmm, I wouldn't worry about it too much, it's definitely nice to be able to pass in your own axes but it's probably a niche requirement here. 

If you really wanted to do it right, I have a feeling that where you need multiple _axes_, what you need is a _figure_. But I don't understand matplotlib very well so that might be wrong, and if it's right I don't know how you'd actually use it.

valschneider:
- Removed unused parameters `title`, `ax` and `legend_label` from cdev plotting methods
- Removed related pydoc
- Removed legend generation for cdev plotting methods since there's only one data set being drawn per plot.

tests/test_thermal.py
```diff
@@ -139,6 +139,46 @@ def test_plot_inout_power(self):
         trappy.FTrace().thermal_governor.plot_inout_power(title="Antutu")
         matplotlib.pyplot.close('all')
 
+class TestCpuOutPower(BaseTestThermal):
+    def __init__(self, *args, **kwargs):
+        super(TestCpuOutPower, self).__init__(*args, **kwargs)
```
bjackman: Don't need this, it's implied

tests/test_thermal.py
```diff
@@ -139,6 +139,46 @@ def test_plot_inout_power(self):
...
+        Can't check that the graph is ok, so just see that the method doesn't blow up"""
+        out = trappy.FTrace().cpu_out_power
+
+        out.plot_cdev_states()
+        out.plot_cdev_states(xlim=(3, 6))
+        matplotlib.pyplot.close('all')
+
+class TestDevfreqOutPower(BaseTestThermal):
+    def __init__(self, *args, **kwargs):
+        super(TestDevfreqOutPower, self).__init__(*args, **kwargs)
```
bjackman: ditto

trappy/cpu_power.py
```diff
@@ -120,6 +143,87 @@ def get_all_freqs(self, mapping_label):
...
+
+        :param height: The height of the plot
+        :type height: int
+
+        :param xlim: The xlim setting of the plot.
+            See :func:`~trappy.plot_utils.set_lim`
+        :type xlim: str or tuple of int
+
+        :param ylim: The ylim setting of the plot
+            See :func:`~trappy.plot_utils.set_lim`
```
bjackman: What does the '~' do?

valschneider: According to this: https://stackoverflow.com/a/21297186/5096023
it's a way to setup a reference to another class' methods. I built the doc in HTML and it did indeed give me a hypertext link to that method's doc.

bjackman: Oh right, I thought it did that automatically ¯\_(ツ)_/¯

valschneider:
Sure thing, I'd welcome some more feedback. I'll just fixup that `__init__` thing.

valschneider:
One more thing: I was thinking of adding plotting methods for the **thermal_power_{cpu/devfreq}_{get/limit}** traces. Thing is, the **thermal_power_allocator** trace has the same data concatenated in a single trace BUT with mangled device names.

I was thinking of just reading the data from **thermal_power_{cpu/devfreq}_{get/limit}** in LISA to populate the `power_actors` argument of `thermal_power_allocator.plot_{in/out}put_power()`, but it feels like it would be a cleaner job to have specific plotting methods in the **devfreq/cpu** related classes. Do you think that's worth to have ?

valschneider:
@bjackman Added a `plot_generic` method in **plot_utils.py**. It allows me to factorize code for cpufreq/devfreq plotting, and also makes it easy to plot stuff on the go: https://gist.github.com/valschneider/89cbcc9318d80af74c4cc3f84cd8e075#file-thermal_plotting-ipynb

derkling:
As an aside comment, there was a discussion in the past about moving plotting functionalities out of TRAPpy which should instead focus on trace parsing and DF generation.

Seems that in the past the main argument against this idea was backward compatibility.

Since here it seems we are at adding yet some other plots, I was wondering if we should not raise that question again and have another discussion about the opportunity to consolidate the tools we have around. To me the overall idea design should be:
 - WA: data collection
 - TRAPpy: data post-processing
 - LISA: data analysis and this plotting

What do you think guys, what can be done to improve on this overall design?

valschneider:
@derkling I was actually surprised to see there were plotting methods in TRAPpy- it's only for the thermal modules though. 

Tbh, I believe it would be indeed cleaner to keep the plotting methods contained in LISA, but I also see the appeal of being able to easily plot stuff from within TRAPpyitself: you don't need yet another repository.

However, there's already a lot of plotting methods in LISA, and only thermal plots in TRAPpy. I'm not leaning strongly towards either choice, but I do believe either one should have plotting methods, not both.

bjackman:
@derkling  yeah these are good points, and I think your proposed split of responsibilities makes good sense. 

The only problem I can see is that LISA cannot currently be used as a library (i.e. it uses the init_env thing) - it would be cool to be able to integrate LISA into other tools (e.g. as a results_processor in WA), which can already be done with TRAPpy. I did a little fiddling to see if that can be fixed, it seems like it definitely could but it would require reorganising the repository, and would also break backward compatibility because all imports would need to be changed (e.g. I think `from trace import Trace` would need to become `from lisa.trace import Trace` or similar). Do you know why we have the init_env thing?

trappy/plot_utils.py
```diff
@@ -140,6 +140,114 @@ def number_freq_plots(runs, map_label):
 
     return num_freq_plots
 
+def plot_generic(dfr, name_field, names=None, fields=None,
+                 prettify_name=None, width=None, height=None, xlim="default",
+                 ylim="range", drawstyle="default", ax=None, title=""):
```
bjackman: So, on the `plot_generic` thing, it's a good idea but it duplicates some logic in trappy, namely the grammar use by `trappy.stats.grammar.Parser` and `trappy.ILinePlot`. 

```
plot_generic(thermal_dfr, "cpus", fields=["cdev_state"])
```

would roughly (I haven't actually tested this) translate to

```
ILinePlot(thermal_dfr, column="cdev_state", pivot="cpus")
```

This actually does something totally different because it only works for the interactive plots that can be used in IPython, not matplotlib. So what we ideally want is probably to extend that existing code to work with matplotlib instead. That might be a bit of a lofty goal (not sure), but we should definitely make the two interfaces use the same terminology and if possible, design the interface so that it _looks like_ the code is shared :P 

What you've called a "determinant" is called a "pivot" in the existing code, and what you've called "names" is implemented with "filters". So if you just want to plot CPU 0's load you do `ILinePlot(sched_load_avg_cpu_df, column='util_avg', pivot='cpu', filters={'cpu': [0]})`.

But this is probably too much detail, we should first finish the discussion started by @derkling about where plotting code belongs, I think.

trappy/plot_utils.py
```diff
@@ -140,6 +140,114 @@ def number_freq_plots(runs, map_label):
...
+        setup_plot = True
+
+    for name in names:
+        cols = []
+        for field in fields:
+            cols.append('{} {}'.format(prettify_name(name), field))
+
+        plot_dfr = dfr[dfr[name_field] == name][fields]
+        plot_dfr.columns = cols
+        plot_dfr.plot(ax=ax, drawstyle=drawstyle)
```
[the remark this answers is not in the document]
valschneider: I went this way because I was getting duplicate index issues that I couldn't fix even with the `handle_duplicate_index()` utility. But since I haven't mastered the Dataframe black magic, there might be yet another way to go around this issue.

valschneider:
Closing this for now as after chatting with @derkling & @bjackman, it really makes more sense to have the plotting utilities in LISA.

derkling:
@bjackman the init_env is there mainly because LISA tracks devlib and trappy as submodules and we want to be sure the local libraries are referenced. This was required at the beginning of the development when the LISA APIs was mostly dependant on leading-edge features of those libs.
Actually, also more recently we usually merge LISA APIs which progress the dependencies as well... while those dependencies are not yet released as stable pip packages.

Finally, we use the init_env to setup the LISAShell which provides a set of custom convenience commands.

Another question is: do we really need to use LISA as a dependency of some other tool?
In my view, LISA is that "umbrella" project which uses devlib, trappy and potentially WA... but none of these tools should be interested in using LISA as a dependency.

bjackman:
(Repeating conclusion of a f2f discussion)

Agreed - my proposed use case for using LISA as a dependency of another tool is not really necessary, I think (or perhaps it could conceivably be useful but it is not worth the cost). Let's try to move plotting and analysis into LISA.

bjackman:
Although.. saying that I've just remembered that TRAPpy has all the interactive plot stuff and "plotting" in the name! I suppose the proposal is to keep the generic stuff (ILinePlot & co) in TRAPpy and move out all the stuff that encodes knowledge about Linux itself?

valschneider:
@bjackman I'll first move my PR & the other interesting plotting methods, and then we can see what remains in TRAPpy and what can be weeded.

PR #270, opened 2017-10-12
Caching: factorizing / toughening

valschneider:
I've been struggling to reproduce the bugs generated by caching (I did encounter them in the past so I'm not doubting their existence), as such I mostly went through the code and improved what I thought needed to be improved.

I'm not re-enabling caching in this PR, but I will try to convince some people to re-enable it on their side and do some bug-finding for me.

### DIFF
--- a/tests/test_caching.py
+++ b/tests/test_caching.py
@@ -14,6 +14,7 @@
 #
 
 import os
+import json
 import shutil
 import sys
 import unittest
@@ -107,24 +108,30 @@ class TestCaching(utils_tests.SetupDirectory):
         src = os.path.join(utils_tests.TESTS_DIRECTORY, "trace_sched.txt.cache")
         shutil.copytree(src, cache_path)
 
-        md5_path = os.path.join(cache_path, "md5sum")
-        def read_md5sum():
-            with open(md5_path) as f:
-                return f.read()
+        metadata_path = os.path.join(cache_path, "metadata.json")
+
+        def read_metadata():
+            with open(metadata_path, "r") as f:
+                return json.load(f)
+
+        def write_md5(md5):
+            metadata = read_metadata()
+            metadata["md5sum"] = md5
+            with open(metadata_path, "w") as f:
+                json.dump(metadata, f)
+
 
         # Change 1 character of the stored checksum
-        md5sum = read_md5sum()
-        # Sorry, I guess modifying strings in Python is kind of awkward?
-        md5sum_inc = "".join(list(md5sum[:-1]) + [chr(ord(md5sum[-1]) + 1)])
-        with open(md5_path, "w") as f:
-            f.write(md5sum_inc)
+        md5sum = read_metadata()["md5sum"]
+        md5sum_inc = md5sum[:-1] + chr(ord(md5sum[-1]) + 1)
+        write_md5(md5sum_inc)
 
         # Parse a trace, this should delete and overwrite the invalidated cache
         GenericFTrace.disable_cache = False
         trace = trappy.FTrace()
 
         # Check that the modified md5sum was overwritten
-        self.assertNotEqual(read_md5sum(), md5sum_inc,
+        self.assertNotEqual(read_metadata()["md5sum"], md5sum_inc,
                             "The invalid ftrace cache wasn't overwritten")
 
     def test_cache_dynamic_events(self):
@@ -200,7 +207,7 @@ class TestCaching(utils_tests.SetupDirectory):
         trace_dir = os.path.dirname(trace_path)
         trace_file = os.path.basename(trace_path)
         cache_dir = '.' + trace_file + '.cache'
-        number_of_trace_categories = 29
+        number_of_trace_categories = 27
         self.assertEquals(len(os.listdir(cache_dir)), number_of_trace_categories)
 
         os.remove(os.path.join(cache_dir, 'SchedWakeup.csv'))
--- a/tests/test_caching.py
+++ b/tests/test_caching.py
@@ -21,38 +21,42 @@ import unittest
 import utils_tests
 import trappy
 from trappy.ftrace import GenericFTrace
+from trappy.systrace import SysTrace
 
 class TestCaching(utils_tests.SetupDirectory):
     def __init__(self, *args, **kwargs):
         super(TestCaching, self).__init__(
             [("trace_sched.txt", "trace.txt"),
-             ("trace_sched.txt", "trace.raw.txt")],
+             ("trace_sched.txt", "trace.raw.txt"),
+             ("trace_systrace.html", "trace.html")],
             *args,
             **kwargs)
 
     def test_cache_created(self):
         """Test cache creation when enabled"""
         GenericFTrace.disable_cache = False
-        trace = trappy.FTrace()
+        traces = (trappy.FTrace(), trappy.SysTrace(path='./trace.html'))
 
-        trace_path = os.path.abspath(trace.trace_path)
-        trace_dir = os.path.dirname(trace_path)
-        trace_file = os.path.basename(trace_path)
-        cache_dir = '.' + trace_file + '.cache'
+        for trace in traces:
+            trace_path = os.path.abspath(trace.trace_path)
+            trace_dir = os.path.dirname(trace_path)
+            trace_file = os.path.basename(trace_path)
+            cache_dir = '.' + trace_file + '.cache'
 
-        self.assertTrue(cache_dir in os.listdir(trace_dir))
+            self.assertTrue(cache_dir in os.listdir(trace_dir))
 
     def test_cache_not_created(self):
         """Test that cache should not be created when disabled """
         GenericFTrace.disable_cache = True
-        trace = trappy.FTrace()
+        traces = (trappy.FTrace(), trappy.SysTrace(path='./trace.html'))
 
-        trace_path = os.path.abspath(trace.trace_path)
-        trace_dir = os.path.dirname(trace_path)
-        trace_file = os.path.basename(trace_path)
-        cache_dir = '.' + trace_file + '.cache'
+        for trace in traces:
+            trace_path = os.path.abspath(trace.trace_path)
+            trace_dir = os.path.dirname(trace_path)
+            trace_file = os.path.basename(trace_path)
+            cache_dir = '.' + trace_file + '.cache'
 
-        self.assertFalse(cache_dir in os.listdir(trace_dir))
+            self.assertFalse(cache_dir in os.listdir(trace_dir))
 
     def test_compare_cached_vs_uncached(self):
         """ Test that the cached and uncached traces are same """
--- a/tests/test_caching.py
+++ b/tests/test_caching.py
@@ -203,6 +203,20 @@ class TestCaching(utils_tests.SetupDirectory):
 
         self.assertEqual(len(trace1.sched_wakeup.data_frame), 2)
 
+    def test_ftrace_metadata(self):
+        """Test that caching keeps trace metadata"""
+        GenericFTrace.disable_cache = False
+
+        self.test_cache_created()
+
+        trace = trappy.FTrace()
+
+        version = int(trace._version)
+        cpus = int(trace._cpus)
+
+        self.assertEquals(version, 6)
+        self.assertEquals(cpus, 6)
+
     def test_cache_delete_single(self):
         GenericFTrace.disable_cache = False
         trace = trappy.FTrace()
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -68,65 +68,6 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
 
     disable_cache = True
 
-    def _trace_cache_path(self):
-        trace_file = self.trace_path
-        cache_dir  = '.' +  os.path.basename(trace_file) + '.cache'
-        tracefile_dir = os.path.dirname(os.path.abspath(trace_file))
-        cache_path = os.path.join(tracefile_dir, cache_dir)
-        return cache_path
-
-    def _check_trace_cache(self, params):
-        cache_path = self._trace_cache_path()
-        trace_metadata_path = os.path.join(cache_path, 'metadata.json')
-
-        if not os.path.exists(trace_metadata_path):
-            return False
-
-        with open(trace_metadata_path) as f:
-            trace_metadata = json.load(f)
-
-        with open(self.trace_path, 'rb') as f:
-            trace_md5sum = hashlib.md5(f.read()).hexdigest()
-
-        self.basetime = float(trace_metadata["basetime"])
-
-        # Check if cache is valid
-        if trace_metadata["md5sum"] != trace_md5sum:
-            warnstr = "Cached data is from another trace, invalidating cache."
-            warnings.warn(warnstr)
-            shutil.rmtree(cache_path)
-            return False
-
-        # Check if cache can be used with given parameters
-        # Convert to a json string for comparison
-        if json.dumps(trace_metadata["params"]) != json.dumps(params):
-            warnstr = "Cached trace parameters differ from those given, invalidating cache."
-            warnings.warn(warnstr)
-            shutil.rmtree(cache_path)
-            return False
-        return True
-
-    def _create_trace_cache(self, params):
-        cache_path = self._trace_cache_path()
-        trace_metadata_path = os.path.join(cache_path, 'metadata.json')
-
-        if os.path.exists(cache_path):
-            shutil.rmtree(cache_path)
-        os.mkdir(cache_path)
-
-        trace_metadata = {}
-
-        trace_metadata["md5sum"] = hashlib.md5(open(self.trace_path, 'rb').read()).hexdigest()
-        trace_metadata["basetime"] = self.basetime
-        trace_metadata["params"] = params
-
-        with open(trace_metadata_path, 'w') as f:
-            json.dump(trace_metadata, f)
-
-    def _get_csv_path(self, trace_class):
-        path = self._trace_cache_path()
-        return os.path.join(path, trace_class.__class__.__name__ + '.csv')
-
     def __init__(self, name="", normalize_time=True, scope="all",
                  events=[], window=(0, None), abs_window=(0, None)):
         super(GenericFTrace, self).__init__(name)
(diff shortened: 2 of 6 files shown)


trappy/ftrace.py
```diff
@@ -97,13 +97,59 @@ def _check_trace_cache(self, params):
...
+            self.abs_window = abs_window
+
+            if before_window or after_window:
+                warnstr = "Cached trace parameters incompatible " + \
+                          "with those given, invalidating cache."
+                warnings.warn(warnstr)
+                shutil.rmtree(cache_path)
+                return False
+
+            self.cache_window_differs = True
```
bjackman: Since we have had correctness issues with this stuff in the past I would not really be comfortable with all this window checking unless there is really good test coverage.. if you have time to add that, then let's go ahead, but otherwise I am more than happy to live with overly-aggressive cache invalidation in return for a little bit more confidence about correctness. After all I don't think the window param even gets used very often..

tests/test_caching.py
```diff
@@ -199,6 +199,28 @@ def test_cache_window(self):
...
+        # Parse again with a shifted window
+        trace2 = trappy.FTrace(
+            events=['sched_wakeup'],
+            window=(1, 2))
+
+        # Cache should have been invalidated to fetch data after first window
+        self.assertGreater(
+            trace2.sched_wakeup.data_frame.index[-1],
+            trace1.sched_wakeup.data_frame.index[-1]
+        )
```
bjackman: Ah, I didn't spot this, it should probably go in the same commit as the functionality it tests. This coverage definitely helps, but as I said I'd rather be really conservative here and only go for the window checking if we have tested multiple corners of the input space.

trappy/ftrace.py
valschneider: I don't use the window parameters that much myself either - I manually trim the dataframes if I have to. Thing is, if someone does use them, this makes quite a big difference in terms of parsing time.

trappy/ftrace.py
```diff
@@ -226,6 +226,14 @@ def _do_parse(self):
                         continue
                     csv_file = self._get_csv_path(trace_class)
                     trace_class.write_csv(csv_file)
+
+                # Remove the text file since we have a cache
+                if self.txt_deletable:
+                    from subprocess import check_output
+
+                    cmd = ["rm", self.trace_path]
+                    check_output(cmd)
```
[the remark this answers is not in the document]
valschneider: I suppose we could. I instored the `txt_deletable` flag to detect whether or not the trace.txt has been created by a user, but maybe there are cases that are not covered ?

trappy/ftrace.py
```diff
@@ -77,51 +77,51 @@ def _trace_cache_path(self):
...
+        # Check if cache is valid
+        if cache_info["md5sum"] != trace_md5sum:
+            warnstr = "Cached data is from another trace, invalidating cache."
+            warnings.warn(warnstr)
+            shutil.rmtree(cache_path)
+            return False
+
+        # Check if cache can be used with given parameters
+        # Convert to a json string for comparison
+        if json.dumps(cache_info["params"]) != json.dumps(params):
```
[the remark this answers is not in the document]
valschneider: It probably is. It was already saved in the previous implementation - it's mostly to quickly restore it, as it's obtained when parsing the trace.txt file.

trappy/ftrace.py
```diff
@@ -193,6 +193,7 @@ def unregister_parser(cls, cobject):
 
     def _do_parse(self):
         params = {'window': self.window, 'abs_window': self.abs_window}
+
```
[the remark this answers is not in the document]
valschneider: Noise will be trialled and executed tomorrow morning.

valschneider:
- Removed noise
- Squashed window test commit into window-aware commit

valschneider:
I'm looking into using a Tempfile instead of a trace.txt whenever possible. Sadly, I realised my "Delete trace.txt if caching is enabled" commit isn't actually working. If the txt file is correctly deleted after the cache is created, it is re-generated even if the cache exists. I can fix this but it requires a fair amount of rework, stay tuned.

valschneider:
`GenericFTrace` has gone through a facelift, and I've broken systrace (I haven't tested it but I'm sure I did), however it looks like there are no systrace tests so I'll add some next week.

This "v2" doesn't have any window check as @bjackman suggested, but it has:
- Shiny new factorized code for `GenericFtrace`
- trace-cmd report outputted to a tempfile that is deleted once parsing is complete

bjackman:
Cool, thanks, I'll start testing this first and then review later.

valschneider:
Surprisingly, I hadn't broken systrace. I did make it parse the file twice though, this is fixed now. I also realized that removing the file extension from the cache dir name wasn't such a smart idea: there could be a trace.dat and trace.html living in the same directory, and they should each get their own cache.

tl;dr:
- Fixed Systrace and added Systrace tests
- Reverted cache directory format from ".trace.cache/" to ".trace.{txt/dat/html}.cache/"

valschneider:
@bjackman: I had a chat with @derkling, we could have a fairly simple handling of the windows. If caching is enabled, we would disregard windows and parse the whole trace to cache it, and then return the requested window. As such, whatever window may be requested later on, no more parsing is required.

bjackman:
OK, makes sense

valschneider:
Pushed that cached-window-handling thing, but for some reason the PR page won't update.

valschneider:
Okay looks like it eventually made its way there.

trappy/ftrace.py
```diff
@@ -676,18 +738,30 @@ def __process_path(self, basepath):
...
+            warnstr = ".raw.txt trace format is no longer supported"
+            raise RuntimeError(warnstr)
+        else:
+            warnstr = "Could not find any trace file in {}".format(basepath)
+            raise IOError(warnstr)
+
+        return trace_to_read
+
+    def __generate_trace_txt(self, trace_dat):
+        return self.__run_trace_cmd_report(trace_dat)
```
bjackman: Let's move the body of `__run_trace_cmd_report` in here, or remove this method and just call `__run_trace_cmd_report` directly.

trappy/ftrace.py
```diff
@@ -191,46 +188,95 @@ def unregister_parser(cls, cobject):
...
+
+    def _load_cache(self):
+        if self._check_trace_cache():
+            self.max_window = self._max_window()
+             # Read csv into frames
             for trace_class in self.trace_classes:
                 try:
                     csv_file = self._get_csv_path(trace_class)
                     trace_class.read_csv(csv_file)
+                    self._windowify_class(trace_class, self.max_window)
```
bjackman: Let's have a comment here explaining that the cache was written without the window applied, so we need to apply it after reading.

trappy/ftrace.py
```diff
@@ -306,14 +360,18 @@ def __populate_data(self, fin, cls_for_unique_word):
 
             if not self.basetime:
                 self.basetime = timestamp
-
-            if (timestamp < self.window[0] + self.basetime) or \
-               (timestamp < self.abs_window[0]):
+                # Disregard windows if updating/creating cache
+                # This way the cache is ready for later window requests
+                if not self.__class__.disable_cache:
```
bjackman: I think this is unecessary complexity: disabled caching is supposed to be a niche feature for testing and debugging, so we don't care about performance in that case. So I think when parsing a trace file we can _always_ ignore the window, and then _always_ apply the window after parsing is finished, using the new `windowify` methods. Does that work?

trappy/ftrace.py
```diff
@@ -306,14 +360,18 @@ def __populate_data(self, fin, cls_for_unique_word):
...
+                    self.max_window = self._max_window()
```
bjackman: IIUC this is set in two separate places depending on whether we're reading from the cache, but it's actually independent of that. Can we just do it in one place at the beginning? Or even make `self.max_window` a `property`?

trappy/ftrace.py
```diff
@@ -273,6 +319,14 @@ def __get_trace_class(self, line, cls_word):
                     return trace_class
         return trace_class
 
+    def __before_window(self, timestamp):
+        return (timestamp < self.window[0] + self.basetime) or \
+               (timestamp < self.abs_window[0])
```
bjackman: Sorry, too lazy to figure this out myself: can this be simplified using the new `max_window` thing?

trappy/ftrace.py
```diff
@@ -69,10 +71,10 @@ class GenericFTrace(BareTrace):
...
 
     def _trace_cache_path(self):
-        trace_file = self.trace_path
-        cache_dir  = '.' +  os.path.basename(trace_file) + '.cache'
-        tracefile_dir = os.path.dirname(os.path.abspath(trace_file))
-        cache_path = os.path.join(tracefile_dir, cache_dir)
+        trace_name = os.path.basename(self.trace_path)
+        trace_dir = os.path.dirname(os.path.abspath(self.trace_path))
+        cache_dir  = '.' + trace_name + '.cache'
+        cache_path = os.path.join(trace_dir, cache_dir)
```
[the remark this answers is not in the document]
valschneider: I think this is a piece of orphaned code that ended up here after a few rewrites. I'll clean that up.

trappy/ftrace.py
```diff
@@ -216,15 +204,42 @@ def _update_cache(self):
...
                 "TRAPpy: Cache not created due to OS error: {0}".format(err))
 
-    def _load_cache(self):
-        params = self._get_params_to_cache()
+    def _max_window(self):
+        """
+        Compute the maximum window of the 'window' & 'abs_window' intersection
+        """
+        max_window = [0, None]
+        max_window[0] = max(self.window[0] + self.basetime, self.abs_window[0])
```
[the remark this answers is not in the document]
valschneider: This is because of how windows are handled in the parsing (and is one of the reasons why I really don't like how windows are handled in TRAPpy, I thought you were right but it took me a while to re-understand what's going on) in `__populate_data`:

```
if (timestamp < self.window[0] + self.basetime) or \
    (timestamp < self.abs_window[0]):
    self.lines += 1
    continue
```

In other words, an event is skipped if its timestamp is lower than *either* bound. This means events will start being considered if their timestamp is greater than the greatest lower bound, so `max` should fit the bill.

trappy/ftrace.py
valschneider: So at first the code was:

```
if (timestamp < self.window[0] + self.basetime) or \
(timestamp < self.abs_window[0]):
    self.lines += 1
    continue

if (self.window[1] and timestamp > self.window[1] + self.basetime) or \
    (self.abs_window[1] and timestamp > self.abs_window[1]):
    return
```

I then turned each `if` condition in a method (`__before_window` and `__after_window`) in the **Methodify window inclusion check** commit. These methods are used in the computation of `max_window`, but it still makes sense to me to keep them as they are used in `__populate_data` for two different checks.

trappy/ftrace.py
valschneider: That's what I wanted to do, sadly it **needs** to be set in two separate places. If we have the cache, we can read the `basetime` parameter from it and create the max window. If not, we must start parsing the file to finally get the `basetime` and then we can create that max window value.

trappy/ftrace.py
valschneider: That would work, but the incentive for doing this with a cache is that we're paying a cost (parsing time) that will later on be cushioned by the use of the cache. If caching is disabled, we should still mind the window so we stop parsing the trace once we're after the window.  I understand disabled cache should become an exotic use-case but I felt like this was still somewhat justified.

bjackman: I see where you're coming from but I think we should make our task as easy as possible by only having one use case that we really optimise, i.e. `cache_disabled=False`

valschneider: Okay, I'll remove that bit.

valschneider:
@bjackman I think I addressed all of yours comments. 

I also made a slight change to `_do_parse`: I factorized the windowification and time normalization in a `_apply_user_parameters` method

valschneider:
Food for thoughts: had a chat with @derkling, following the "disregard window" initiative we could also parse all possible events to store them in the cache, so we would only need to cache a trace once, no matter the parameters. This might be a bit more extreme than to simply disregarding windows, so it might be better to have it in another PR - but it's a thing to consider, at least.

trappy/ftrace.py
```diff
@@ -306,14 +367,17 @@ def __populate_data(self, fin, cls_for_unique_word):
 
             if not self.basetime:
                 self.basetime = timestamp
+                # Disregard windows, this way the cache will be ready for later
+                # window requests
+                self.max_window = self._max_window()
+                self.window = (0, None)
+                self.abs_window = (0, None)
```
bjackman: Instead of setting a window that does nothing here, why don't we just delete the __before/after_window checks (and methods themselves) below? 

AFAICT `self.window` and `self.abs_window` are no longer used except to calculate `self.max_window`.

trappy/ftrace.py
```diff
@@ -644,28 +708,33 @@ class FTrace(GenericFTrace):
...
+    def _parsing_setup(self):
+        super(FTrace, self)._parsing_setup()
+
+        if self.read_from_dat:
+            self.file_to_parse = self.__generate_trace_txt(self.trace_path)
+
+        # file_to_parse is already set to trace_path in the superclass,
+        # so no "else" is needed here
+
         self.__populate_metadata()
```
bjackman: I think it would be clearer to put `self.__populate_metadata` at the end of `__init__` directly, putting it here gives a false impression that it needs to be done at this specific stage of parsing.

trappy/ftrace.py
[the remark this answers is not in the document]
valschneider: Probably never happens because we don't use windows most of the time, and if we do it's either `window` or `abs_window`, not both.

trappy/ftrace.py
valschneider: SysTrace is also a subclass of GenericFtrace but does not have any `__populate_metadata` method, which is why I put it here.

bjackman:
> Food for thoughts: had a chat with @derkling, following the "disregard window" initiative we could also parse all possible events to store them in the cache, so we would only need to cache a trace once, no matter the parameters. This might be a bit more extreme than to simply disregard windows, so it might be better to have it in another PR - but it's a thing to consider, at least.

Agreed, e

trappy/ftrace.py
valschneider: If we want to truly commit to having cache enabled by default, I suppose we could, yes.

trappy/ftrace.py
valschneider: Wait no, the `__{before/after}_window` methods are no longer used in the computation of `max_window`, that was in a previous version.

bjackman:
Sorry, that "Close and comment" button always looks so much like a "cancel posting this comment".

> Food for thoughts: had a chat with @derkling, following the "disregard window" initiative we could also parse all possible events to store them in the cache, so we would only need to cache a trace once, no matter the parameters. This might be a bit more extreme than to simply disregard windows, so it might be better to have it in another PR - but it's a thing to consider, at least.

Basically yeah, agree.

valschneider:
Can't seem to reply directly to 2 of your comments, replying here:

>Huh, so an event is dropped if it's earlier than either window[0]s or later than both window[1]s? That doesn't sound good. But OK, let's leave it for another day.

We never encounter this because we either use no window, or just one (never both at the same time). But yes, it's a bit weird.

valschneider:
Removed the window checks altogether from `_do_parse()`

trappy/ftrace.py
bjackman: But can't it just go at the end of `FTrace.__init__`

valschneider: Woops, good thing you told me to look at that. Turns out I hadn't updated `__populate_metadata` to use `file_to_parse` instead of `trace_path`, so it wasn't doing anything actually.

Also, yes, you're right, it can be moved there. I thought you were talking about the superclass.

valschneider: Argh ! Metadata aren't being saved in the cache ! This means we were missing out things like trace._cpus, which would explain some of the issues we had with cache. Thanks for pointing me in the right direction, I got some more work to do...

valschneider:
Alright, we're getting there - thanks a lot @bjackman for the careful review.

I've (once again) re-arranged the commits and did some pedantic changes here and there, but it's roughly what it was before with the following changes:

- Fiddling with trace metadata made me realise what I called 'cache_info' is really just metadata, so the file is now named 'metadata.json' and it makes more sense to me.
- Extra data can be appended in to this file with by overriding the `_get_extra_data_to_cache` and `_load_extra_data_from_cache` methods. In the case of FTrace, this is used to store trace metadata (lines of data that appear in the trace but are not actual trace events).
- Added cached/uncached FTrace metadata test

> I think it would be clearer to put self.__populate_metadata at the end of __init__ directly, putting it here gives a false impression that it needs to be done at this specific stage of parsing.

Right now it has to be in there, otherwise metadata will not be cached.

valschneider:
Fixed an annoying warning that said cache was invalid when it simply hadn't been created yet.

trappy/ftrace.py
```diff
@@ -173,74 +173,113 @@ def _trace_cache_path(self):
...
+                # Write cache metadata
+                metadata_path = os.path.join(self._trace_cache_path(), 'metadata.json')
 
-            # Write out only events that weren't cached before
+                metadata = {}
+                metadata["md5sum"] = hashlib.md5(
+                    open(self.trace_path, 'rb').read()
+                ).hexdigest()
+                metadata["basetime"] = self.basetime
+                metadata["extra"] = self._get_extra_data_to_cache()
```
bjackman: Instead of having an "extra" key in `metadata` and `_get_extra_data_to_cache` and `_load_extra_data_from_cache`, have you considered just having `_get_metadata_to_cache` and `_load_metadata_from_cache`, then having the subclasses use `super` then extend the result?

So, `GenericFTrace._get_metadata_to_cache` just returns `{"md5sum": <hash>,  "basetime": self.basetime"}`, then `FTrace` overrides it like:

```python
def _get_metadata_to_cache(self):
    ret = super(FTrace, self)._get_metadata_to_cache()
    ret['_version'] = self._version
    ret['_cpus'] = self._cpus
    return ret
```

Then `GenericFTrace._load_metadata_from_cache` (called before `_is_cache_valid`) just sets `self.basetime`, and `FTrace` overrides it like:

```python
def _load_metadata_from_cache(self, metadata):
    ret = super(FTrace, self)._load_metadata_from_cache(metadata):
    self._cpus = metadata['_cpus']
    self._version = metadata['_version']
```

It feels more in the noble spirit of OOP, rather than drawing a line in the inheritance tree after which everything is 'extra'. Do you think that's clearer?

bjackman: PS one benefit of writing it this way is you can grep for "self._cpus =" when debugging. That's orthogonal to the inheritance design though.

valschneider: I did think about that - it would have worked the same way as `_parsing_setup` & `_parsing_teardown`.

The thing is, `file_to_parse` is set in  `_parsing_setup` but it is **meant** to be changed. So you can override it and not even call the super method, and it's fine.
You could break `_get_metadata_to_cache` by not calling the super method (though that'd just be stupid) OR by writing over pre-existing metadata values (unlikely, but I wanted to play safe). The idea of that "extra" key is to provide a "safe" dict where any key can be used without compromising the required base metadata.

trappy/ftrace.py
```diff
@@ -138,14 +139,41 @@ def unregister_parser(cls, cobject):
             if cobject == obj:
                 del scope_classes[name]
 
+    def _max_window(self):
```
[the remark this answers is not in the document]
valschneider: Oh yes this is a terrible name, I'll change it.

trappy/ftrace.py
```diff
@@ -173,74 +173,113 @@ def _trace_cache_path(self):
...
+
+        metadata = self._get_cache_metadata()
+
+        if not self._is_cache_valid(metadata):
+            shutil.rmtree(cache_path)
+            return
+
+        # Load metadata
+        self.basetime = metadata["basetime"]
+        self.max_window = self._max_window()
```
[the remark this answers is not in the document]
valschneider: If we're not parsing, it has to be set in `__populate_data`. As such, I would like to keep a "symmetry", in that it is not done in `_do_parse` but in each appropriate submethod. Otherwise it feels a bit confusing to me: in the scope of `_do_parse`, a newcomer would see `max_window` is only being sent after the cache is loaded (though that is my own opinion).

valschneider:
- s/_max_window()/_calc_max_window()/g
- Fixed the merge conflict with #271

trappy/ftrace.py
valschneider: Arguably this "extra" dict could be introduced by whatever subclass is overriding this method... So if you're okay with it we could have the metadata handling be done in overridable methods as you suggested.

bjackman: As I considered your points I became more sure that `super` is better :grin: 

> You could break _get_metadata_to_cache by not calling the super method (..) OR by writing over pre-existing metadata values (..).  The idea of that "extra" key is to provide a "safe" dict where any key can be used without compromising the required base metadata.

I guess this is a question of how to do coupling between parent and child classes - should the child be aware of the parent (must call the super method and be aware of keys it shouldn't touch) or the parent be aware of the child (must define and call an "extra" method and write an "extra" key). `super` is a language feature and all child classes must be aware of their parents anyway, so I think the former is more harmonious.

On the other hand, the thing about the child having to know which keys it mustn't touch is a good point. I'm not sure how much it matters when the classes are this tightly coupled anyway, though (they live in the same file). 

Anyway, this is probably getting a bit too philosophical so I'll leave the decision to you :)

trappy/ftrace.py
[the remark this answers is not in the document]
bjackman: Ahh yep, that does make sense actually.

trappy/ftrace.py
valschneider: I think that all things considered, the overriding of a base `_get/load_metadata_from/to_cache` is the less confusing way to go. I'll still use an "extra" key, but defined in the child.

valschneider:
I think I've addressed all of your comments:
- Fixed the metadata caching test
- Turned the caching metadata methods into proper OOP citizens

valschneider:
- Fixed merge conflict with #272

valschneider:
Undesired whitespaces have been trialled and [sentenced](https://www.youtube.com/watch?v=TM_6tgVJb_o).

PR #271, opened 2017-10-19
parse traces from the common clock infrastructure

kdub:
Traces from clock_set_rate, clock_enable, clock_disable need
specialized parsing. Add additional feature to parse these traces.

Test: run the 3 new unit tests (nosetests tests/test_common_clk.py)

Change-Id: Ib93b72697fc4d5eb30cffb914bbe0cb4c4cd872d

### DIFF
--- a/trappy/common_clk.py
+++ b/trappy/common_clk.py
@@ -1,0 +1,58 @@
+#    Copyright 2017 Google, ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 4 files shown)


trappy/common_clk.py
```diff
@@ -0,0 +1,64 @@
...
+from trappy.base import Base
+from trappy.dynamic import register_ftrace_parser, register_dynamic_ftrace
+
+#clock traces are of the form "clk_name field0=x field1=y ..."
+def parse_common_clock_data(data_str, fn):
+    data_l = data_str.split(' ', 1)
+    p = fn.generate_data_dict(data_l[1])
+    p.update({"clk_name" : data_l[0]})
+    return p
+
```
bjackman: I think you can simplify this: if you add an intermediate common class then you only need to write generate_data_dict once. And then you can actually move the `parse_common_clock_data` inside that method. Something like this (untested):

```python
class CommonClkBase(Base):
    def generate_data_dict(self, data_str):
        #clock traces are of the form "clk_name field0=x field1=y ..."
        clk_name, fields = data_str.split(' ', 1)
        ret = super(CommonClkBase, self).generate_data_dict(fields)
        ret['clk_name'] = clk_name
        return ret
```

kdub:
@bjackman Thanks for the review, adopted all suggestions

bjackman:
Welcome aboard :grin:

joelagnel:
Awesome. :):)

PR #273, opened 2017-10-31
ftrace: Homogenize timestamps conversions

valschneider:
By default, the str to float conversion done when reading from csv is
different from the one used when reading from the trace.txt file.

Here's an example:
- trace.txt string timestamps:
  [76.402065, 80.402065, 80.001337]
- parsed dataframe timestamps:
  [76.402065000000007, 80.402065000000007, 82.001337000000007]

- csv string timestamps:
  [76.402065, 80.402065, 80.001337]
- cached dataframe timestamps:
  [76.402064999999993, 80.402064999999993, 82.001337000000007]

To fix this, the timestamps read from the cache are rounded
to microsecond precision, which results in cache-read timestamps
being identical to trace-read timestamps.

Tests have also been added/tweaked to ensure this stays true.

### DIFF
--- a/tests/test_caching.py
+++ b/tests/test_caching.py
@@ -70,27 +70,28 @@ class TestCaching(utils_tests.SetupDirectory):
         cached_trace = trappy.FTrace(uncached_trace.trace_path)
         cached_dfr = cached_trace.sched_wakeup.data_frame
 
-        # Test whether timestamps are the same:
-        # The cached/uncached versions of the timestamps are slightly
-        # different due to floating point precision errors due to converting
-        # back and forth CSV and DataFrame. For all purposes this is not relevant
-        # since such rounding doesn't effect the end result.
-        # Here's an example of the error, the actual normalized time when
-        # calculated by hand is 0.081489, however following is what's stored
-        # in the CSV for sched_wakeup events in this trace.
-        # When converting the index to strings (and also what's in the CSV)
-        # cached: ['0.0814890000001', '1.981491']
-        # uncached: ['0.0814890000001', '1.981491']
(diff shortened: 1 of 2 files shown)


derkling:
Mmm... I think I probably added that comment, isn't it?
However, it happened sometimes in the past that, depending on how you configure the clock source, you can have traces with [ns] timestamps... but AFAIR, that comment should come with a patch which goal was to ensure that we always store the index in [s].[us] format, isn't it?

valschneider:
Yes indeed, I was mostly thinking it could be worth adding a test for this scenario at some point.

valschneider:
As @bjackman aptly pointed out, the origin of the problem is simply that the CSV lookup uses a different floating conversion method as the trace.txt parser. This can be fixed in a much cleaner way by simply forcing `read_csv()` to use python's float() conversion method instead of panda/numpy's. This revised commit does just that.

tests/test_caching.py
```diff
@@ -70,27 +70,27 @@ def test_compare_cached_vs_uncached(self):
...
-        # and compare them below.
+        # - csv string timestamps:
+        #   [76.402065, 80.402065, 80.001337]
+        # - cached dataframe timestamps:
+        #   [76.402064999999993, 80.402064999999993, 82.001337000000007]
+        #
+        # To fix this, the timestamps read from the cache are rounded
+        # to microsecond precision, which results in cache-read timestamps
+        # being identical to trace-read timestamps. This test ensures that
+        # this stays true.
```
bjackman: This comment is out of date now

valschneider: I updated it but forgot to add it to the commit, I'll push it right now

PR #275, opened 2017-11-06
ftrace: Fix parsing textual traces with no trailing line

bjackman:
We currently strip off the last character under the assumption that it is a
'\n'. However it isn't always. So use rstrip to do that instead.


valschneider:
Didn't test but LGTM

PR #276, opened 2017-11-08
trappy: Add support for 'print:' ftrace fallback event

bjackman:
This event is just like tracing_mark_write but presents a small difficulty in
that its name is a Python keyword. This is solved by appending '_' to the
atribute name of the FTrace object.

### DIFF
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -356,7 +356,8 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
         for event_name in events:
             for cls in known_events.itervalues():
                 if (event_name == cls.unique_word) or \
-                   (event_name + ":" == cls.unique_word):
+                   (event_name + ":" == cls.unique_word) or \
+                   (event_name == cls.name):
                     self.class_definitions[cls.name] = cls
                     break
             else:
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -59,6 +59,21 @@ SPECIAL_FIELDS_RE = re.compile(
                         r"(?P<timestamp>[0-9]+(?P<us>\.[0-9]+)?): (\w+:\s+)+(?P<data>.+)"
 )
 
+# This is just a helper to enable printing a helpful warning.
+# See GenericFTrace.__add_warnings
+class _UniqueNameWarner(object):
+    def __init__(self, owner, name, unique_word):
+        self.owner = owner
+        self.name = name
+        self.unique_word = unique_word
+
+    def __getattr__(self, attr):
+        raise AttributeError(
+            'You are trying to access "{}.{}.{}", instead you should '
+            'access "{}.{}.{}". A bug in TRAPpy used to make this work '
+            'non-deterministically, but that has now been fixed.'.format(
+                self.owner, self.unique_word, attr, self.owner, self.name, attr))
+
 class GenericFTrace(BareTrace):
     """Generic class to parse output of FTrace.  This class is meant to be
 subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
@@ -310,6 +325,8 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
 
         self._parsing_teardown()
 
+        self.__add_warnings()
+
     def _parsing_setup(self):
         # By default, the file pointed by trace_path is parsed. However, an
         # intermediate file could be required. Subclasses can override this
@@ -464,6 +481,29 @@ is part of the trace.
         except FTraceParseError as e:
             raise ValueError('Failed to parse ftrace file {}:\n{}'.format(
                 trace_file, str(e)))
+
+    def __add_warnings(self):
+        # We used to have a bug where when you had a known event whose 'name'
+        # attribute != its 'unique_word', and you specified it explicitly in the
+        # 'events' param, we had two attributes - one for the 'name' and one for
+        # the 'unique_word', and it was undefined which would be populated.
+        # Now we just have the attribute from the 'name'. If there is any code
+        # out there that was relying on this bug (i.e. accessing the
+        # 'unique_word' atribute), this will tell them what they need to do to
+        # fix their code.
+
+        for name, cls in self.class_definitions.iteritems():
+            unique_word = cls.unique_word.split(':')[0]
+            if name == unique_word:
+                continue
+
+            # Sanity check: if we still have both attributes, something went
+            # wrong.
+            if hasattr(self, unique_word):
+                raise RuntimeError(
+                    'I have both "{}" and "{}" attrs '.format(name, unique_word))
+
+            setattr(self, unique_word, _UniqueNameWarner(self, name, unique_word))
 
     # TODO: Move thermal specific functionality
 
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -445,6 +445,12 @@ is part of the trace.
                 continue
 
             unique_word = trace_class.unique_word
+            if unique_word in cls_for_unique_word:
+                # This means TRAPpy has a nasty bug, like the one fixed in
+                # https://github.com/ARM-software/trappy/pull/276
+                raise RuntimeError('Found two parsers for unique word "{}" ({}, {})'
+                                   .format(unique_word, trace_class,
+                                           cls_for_unique_word[unique_word]))
             cls_for_unique_word[unique_word] = trace_class
 
         if len(cls_for_unique_word) == 0:
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -340,7 +340,7 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
             for cls in known_events.itervalues():
                 if (event_name == cls.unique_word) or \
                    (event_name + ":" == cls.unique_word):
-                    self.class_definitions[event_name] = cls
+                    self.class_definitions[cls.name] = cls
                     break
             else:
                 kwords = {
--- a/tests/test_caching.py
+++ b/tests/test_caching.py
@@ -248,7 +248,7 @@ class TestCaching(utils_tests.SetupDirectory):
         trace_dir = os.path.dirname(trace_path)
         trace_file = os.path.basename(trace_path)
         cache_dir = '.' + trace_file + '.cache'
-        number_of_trace_categories = 31
+        number_of_trace_categories = 32
         self.assertEquals(len(os.listdir(cache_dir)), number_of_trace_categories)
 
         os.remove(os.path.join(cache_dir, 'SchedWakeup.csv'))
--- a/tests/test_fallback.py
+++ b/tests/test_fallback.py
@@ -1,0 +1,62 @@
+
+#    Copyright 2017 ARM Limited, Google and contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import trappy
+from utils_tests import SetupDirectory
+
+TEST_DATA = """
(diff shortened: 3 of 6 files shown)


bjackman:
Agh, this is totally broken, I must have been running tests wrong somehow.

bjackman:
Welp.. that was fun :grin: I think I finally fixed it.

valschneider:
Trying to make sure I understand this well enough - please correct me if I'm wrong on any point:

When a trace is created with trace = trappy.trace(...), each trace class dataframe can be accessed by `trace.<trace_class.name>`. However, those dataframes are populated when lines being parsed match the `unique_word` class parameter.

Then there's the whole biz with how trace classes are imported. This is done in:

```
# [2] minus thermal/sched/etc class_definitions.update() done later
self.class_definitions.update(self.dynamic_classes.items())
# [3]
self.__add_events(listify(events))
```

`class_definitions.update()` will update `class_definitions` with keys created from trace class `name` fields, whereas `__add_events()` will update `class_definitions` with keys created from trace class `unique_word` fields.
If that is true (i.e. if I understood things correctly), this means that there is a bug here that has lived comfortably for quite some time but seemingly never bothered us: `class_definitions` can have two different keys pointing to two different instances of the same trace class. In the end, only the entry pointed by the `unique_word` key will be used and have its Dataframe populated.
I hope I'm right, it took me way too much time and too many open browser tabs to write this part.

To me we should've only used the `unique_word` field to index these classes, but then we get events like "print" that cause Python syntax woes... And on top of that, indexing Dataframes by `unique_word` instead of `name` would be a sudden change, that would've had to be done a while ago. So here I think it's mostly a matter of making sure the golden standard used everywhere within the code is indexing by the `name` field.

valschneider:
Hang on, I'm getting a bit confused. According to that:

```
def __parse_trace_file(self, trace_file):
        """parse the trace and create a pandas DataFrame"""

        # Memoize the unique words to speed up parsing the trace file
        cls_for_unique_word = {}
        for trace_name in self.class_definitions.iterkeys():
            trace_class = getattr(self, trace_name)
            if trace_class.cached:
                continue

            unique_word = trace_class.unique_word
            cls_for_unique_word[unique_word] = trace_class

        if len(cls_for_unique_word) == 0:
            return
        ...
```
We actually go through every single instance available in `class_definitions`, BUT we can find two instances of the same trace_class. This means that `cls_for_unique_word` (which is used during parsing to dump actual trace data in the right place) will point to ONE of these two instances, but with dict key iteration being what it is we aren't really guaranteed which version (created during [2] or [3]) we're going to populate. HOWEVER if this is really true I'm surprised we haven't hit this sooner... Am I just completely way off ?

trappy/sched.py
```diff
@@ -66,9 +66,9 @@ def get_pids(self, key=""):
                                           pivot="cpu")
 """Load and Utilization Signals for CPUs"""
 
-SchedContribScaleFactor = register_dynamic_ftrace("SchedContribScaleFactor",
-                                                  "sched_contrib_scale_f:",
-                                                  "sched")
+SchedContribScaleF = register_dynamic_ftrace("SchedContribScaleF",
```
valschneider: Is this change done to make `SchedContribScaleF`'s `name` field the same as its `unique_word` field ? If so, why ? Is that actually required ? There are quite a few classes with different `name` and `unique_word` - at least most if not all of the thermal classes do (see [temperature](https://github.com/ARM-software/trappy/blob/master/trappy/thermal.py#L32) or [devfreq cooling stuff](https://github.com/ARM-software/trappy/blob/master/trappy/devfreq_power.py#L30).

bjackman:
Your first comment matches my understanding exactly.

In your second comment:

> We actually go through every single instance available in class_definitions, BUT we can find two instances of the same trace_class. This means that cls_for_unique_word (which is used during parsing to dump actual trace data in the right place) will point to ONE of these two instances, but with dict key iteration being what it is we aren't really guaranteed which version (created during [2] or [3]) we're going to populate. HOWEVER if this is really true I'm surprised we haven't hit this sooner... Am I just completely way off ?

Yes, this is exactly right, however (as is only just now becoming fully clear to me), we only actually hit the bug when 

- The `events` parameter is used (otherwise [3] never happens) 
- The `events` parameter specifies a defined event class with a different `name` and `unique_word`
- The class is named in `events` by its `unique_word` (otherwise, in [3] the addition to `class_definitions` is a nop as it should be).

The only places where I have seen this happen are

- In `test_systrace.py` where the event is `name='sched_contrib_scale_factor`, `unique_word='sched_contrib_scale_f:'` and the `events` param includes `sched_contrib_scale_f` (note that [the `':'` doesn't break equivalence](https://github.com/ARM-software/trappy/blob/master/trappy/ftrace.py#L342))

- In the test I wrote for this PR where the event is `name='print_'`, `unique_word='print:'` and `events` includes `'print'`

So (replying to your inline comment here)...

> Is this change done to make SchedContribScaleF's name field the same as its unique_word field ?

Not quite. Before this PR, if you have `'sched_contrib_scale_f'` in your `events` param then you will have two `SchedContribScaleFactor` instances, one at `trace.sched_contrib_scale_f` and one at `trace.sched_contrib_scale_factor`, and it's undefined which one will be populated. In practice it seems (because the `TestLegacySystrace` test was passing) that `sched_contrib_scale_f` would be populated. With this PR one of those keys had to go away - by "default" it would be `trace.sched_contrib_scale_f`, my intention here was that the `trace.sched_contrib_scale_factor` should go away instead.

However, on reflection, this should be the other way around - we should change the behaviour for the more specific case (where you specified  `'sched_contrib_scale_f'` in your `events` param) rather than the more general one. I will change this, and I will investigate adding a warning for when you try to access the wrong attribute.

> There are quite a few classes with different name and unique_word - at least most if not all of the thermal classes do (see temperature or devfreq cooling stuff).

I will look into these too.

Thanks a lot for the careful review, this is turning out to be a gnarly one!

bjackman:
Well I've added some more sanity checks and fixed another subtle bug which we have probably never encountered.

My main concern now is whether this breaks any LISA code..

valschneider:
> With this PR one of those keys had to go away - by "default" it would be trace.sched_contrib_scale_f, my intention here was that the trace.sched_contrib_scale_factor should go away instead.

>However, on reflection, this should be the other way around - we should change the behaviour for the more specific case (where you specified 'sched_contrib_scale_f' in your events param) rather than the more general one. I will change this, and I will investigate adding a warning for when you try to access the wrong attribute.

Keeping the dataframes named after `trace_class.name` makes the most sense to me, which is why I was a bit confused as to why you were changing the SchedContribScaleFactor class name.

trappy/ftrace.py
```diff
@@ -356,7 +356,8 @@ class definitions list.  Otherwise, register a class to parse
         for event_name in events:
             for cls in known_events.itervalues():
                 if (event_name == cls.unique_word) or \
-                   (event_name + ":" == cls.unique_word):
+                   (event_name + ":" == cls.unique_word) or \
+                   (event_name == cls.name):
```
valschneider: Personal opinion here - shouldn't we chose between **either** `unique_word` and `name` and stick with it ? This change means we can populate the `events` argument with both trace class names **and** trace event names, and we end up with dataframes named only after trace class names.

Moving to using trace class names as a "unified standard" will break existing code though, so this might not be the best thing to do. Your thoughts ?

trappy/ftrace.py
```diff
@@ -445,6 +445,12 @@ def __parse_trace_file(self, trace_file):
                 continue
 
             unique_word = trace_class.unique_word
+            if unique_word in cls_for_unique_word:
```
valschneider: Admittedly, this PR should fix the bug - can this be verified in tests rather than the actual ftrace parsing code ? This seems more logical to me, but again, personal opinion.

trappy/ftrace.py
```diff
@@ -465,6 +482,38 @@ def __parse_trace_file(self, trace_file):
...
+        # attribute != its 'unique_word', and you specified it explicitly in the
+        # 'events' param, we had two attributes - one for the 'name' and one for
+        # the 'unique_word', and it was undefined which would be populated.
+        # Now we just have the attribute from the 'name'. If there is any code
+        # out there that was relying on this bug (i.e. accessing the
+        # 'unique_word' atribute), this will tell them what they need to do to
+        # fix their code.
+        for name, cls in self.class_definitions.iteritems():
+            # First do some sanity checking (for bugs in TRAPpy)
+            if name != cls.name:
```
valschneider: Can this one actually happen ?

trappy/ftrace.py
bjackman: Hmm yeah I think you make a good point. The problem is that if you get it wrong and use the `name` instead of `unique_word`, the symptom is that you implicitly register a dynamic parser with a default implementation (given how long this took me to figure out, I think that would be completely undebuggable to someone who isn't quite intimate with TRAPpy). That's why I went for the 'accept both' solution here (where the symptom is impossible to debug) and the 'only use `unique_word` solution in the attribute name (where the symptom is an exception with a hopefully-useful message).

However, the `warnings` module lets us warn someone pretty loudly so maybe we can solve it that way... let me try it out.

trappy/ftrace.py
bjackman: Yeah actually you're right, tests are the place to do this. There may have been a reason why I didn't think that was possible.. will investigate..

trappy/ftrace.py
bjackman: No, this is another "should probably be in a test" thing. Will think about it.

bjackman:
> Keeping the dataframes named after trace_class.name makes the most sense to me, which is why I was a bit confused as to why you were changing the SchedContribScaleFactor class name.

I think I disagree with you on this: using the `name` of a trace parser does feel semantically "better" but it also can't be determined except by fiddling with TRAPpy - I think having our own name for an event where it isn't necessary (e.g. `sched_contrib_scale_factor` instead of `sched_contrib_scale_f`) violates the principle of least surprise.

trappy/ftrace.py
bjackman: Hmm, actually the reason I did this as runtime sanity-checks instead of tests was because this is a transient condition - you can't directly detect it in a test because it arises and then disappears during the  constructor of the trace object. I did add a related test for the dynamic parser overwriting a custom one, but if that failed I don't think it will immediately be obvious why.. so this RuntimeError should help.. 

I can still remove it if you prefer though, given that I Have Fixed The Bug Now™

bjackman:
OK on further discussion and examination of existing TRAPpy APIs I changed my mind. So to summarize:

- You can now only access a trace event by its `name`. If you use the `unique_word` instead you get a helpful exception.
- Slightly awkwardly, at construction time you can request it, using the `events` parameter, via either its `name` or its `unique_word`.

e.g. for the sched_contrib_scale_f the `name` is `sched_contrib_scale_factor` and the unique_word is `sched_contrib_scale_f:`. If you construct a trace with `events=['sched_contrib_scale_f']` it will parse it for you, but then if you try to access `trace.sched_contrib_scale_f` you will get an error telling you to access `trace.sched_contrib_scale_factor` instead.

I think this is probably a reasonable balance of semantic cleanliness and usability. Otherwise you'd have to request the `print` event with `events=['print_']` which is a bit ugly.

So.. I'm gona go ahead with this..

trappy/ftrace.py
```diff
@@ -59,6 +59,21 @@ def _plot_freq_hists(allfreqs, what, axis, title):
...
 
+# This is just a helper to enable printing a helpful warning.
+# See GenericFTrace.__add_warnings
+class _UniqueNameWarner(object):
+    def __init__(self, owner, name, unique_word):
+        self.owner = owner
+        self.name = name
+        self.unique_word = unique_word
+
+    def __getattr__(self, attr):
```
douglas-raillard-arm: EDIT: this comment becomes void according to the next one, since that class can disappear altogether.
`__get__` could be used instead, so this object becomes a non-data descriptor and simply accessing it as an attribute will raise an exception:
```
# Currently that raises an exception:
owner.name.data_frame
# But that does not:
owner.name

# with __get__, this would raise an exception as well:
owner.name
```

trappy/ftrace.py
```diff
@@ -465,6 +482,29 @@ def __parse_trace_file(self, trace_file):
             raise ValueError('Failed to parse ftrace file {}:\n{}'.format(
                 trace_file, str(e)))
 
+    def __add_warnings(self):
```
douglas-raillard-arm: Instead of actually adding attributes and creating a placeholder class, implementing `__getattr__` and raising an exception is enough. `__getattr__` is used at the end of the usual lookup sequence, so it will not impact the time it takes to access other attributes.

JaviMerino:
Superseeded by #280, merged in 61d032065ebb

PR #277, opened 2018-04-25
ftrace: fix windowing KeyErrors

derkling:
For some traces and events (e.g. SchedSwitch) we get a KeyError in case
the first event index is bigger then window[0]. For example:

   trace_class:  <trappy.sched.SchedSwitch object at 0x7f3afe425f10>
   window:  [11793.005501, None]
                      __comm  __cpu  __line  __pid       next_comm  next_pid [...]
   Time
   11793.005507        <...>      0       2   6803  shell svc 6799      6800 [...]
   11793.005565        <...>      0       4   6800            adbd      2648 [...]
   11793.005701         adbd      0       7   2648     ->transport      2650 [...]
   11793.005782  ->transport      0      11   2650       trace-cmd      6803 [...]
   11793.005831        <...>      0      17   6803   kworker/u17:0       367 [...]

   [...]

   KeyError: 11793.005501

Where the min window value [11793.005501] is defined by the first valid event
in the trace. Oddly this does not always happens, expecially it seems to not
happen with trappy.dynamic events.

However, since in these cases the dataframe should not be sliced at all,
let's add some boundary checks to ensure we always provide "valid" slicing
index values. This has been verified to fix the above slicing issue in some
test traces.

### DIFF
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -157,13 +157,15 @@ subclassed by FTrace (for parsing FTrace coming from trace-cmd) and SysTrace."""
     def _windowify_class(self, trace_class, window):
         if len(trace_class.data_frame) < 1:
             return
-
+        # Get events DF boundaries
+        min_t = trace_class.data_frame.index[0]
+        max_t = trace_class.data_frame.index[-1]
+        # Restrict events DF boundaries according to the specified window
+        if window[0]:
+            max_t = min(max_t, window[1])
         if window[1]:
-            trace_class.data_frame = trace_class.data_frame[
-                window[0]:window[1]]
-        elif window[0]:
-            trace_class.data_frame = trace_class.data_frame[
-                window[0]:]
+            min_t = max(min_t, window[0])
+        trace_class.data_frame = trace_class.data_frame[min_t:max_t]
 
     def _trace_cache_path(self):
         trace_file = self.trace_path


valschneider:
I need to take some time to have a better look at this - in the meantime, could you add a test (`trappy/tests`) that highlights this issue ? Thanks.

valschneider:
I don't have anything against the patch, the thing is Pandas should be able to handle this:

```python
data = [
    42,
    43,
    55
]

index = [
    10.2,
    10.5,
    10.8
]

df = pd.DataFrame(data, index=index)

print df[10.1:]
print df[10.2:]
print df[10.3:]
```
Outputs:
```
       0
10.2  42
10.5  43
10.8  55

       0
10.2  42
10.5  43
10.8  55

       0
10.5  43
10.8  55

```

@derkling  had encountered this problem a few months back, and AFAIR it was related to this:
https://github.com/pandas-dev/pandas/issues/6066
IOW, his dataframe wasn't sorted which would cause errors during the slice. I can't find it now but I remember reading that the handling of FloatIndex slicing was changed in a more recent pandas version, and after he updated his Pandas it seemingly fixed it.

I think this is a nastier issue than just slicing outside of a dataframe boundary, which is why I'd like to
a) get to the bottom of the bug.
b) have a test that verifies it doesn't happen again if the fix is non-obvious/hacky.

valschneider:
That's a big difference in SciPy versions. Do you get the exception with or without the venv ?

As for making slicing more robust, I don't entirely disagree. What I'm not too keen on is adding code to support library versions that are not used within the venv. IMHO it kind of defeats the purpose of the venv.

Otherwise, if the error shows up in the venv then we should fix the libraries used, and if the error persists (e.g. the expected behaviour was only there in older versions), then it would make sense to change the code.

Admittedly this is me being stubborn and saying:

> why bother adding code to support libraries versions that do not exhibit the expected behaviour
> when we can just tell the user to use the venv / the required library versions

while you're saying

> why bother with the venv when you can fix the problem in a few lines

trappy/ftrace.py
```diff
@@ -157,13 +157,15 @@ def _calc_max_window(self):
     def _windowify_class(self, trace_class, window):
         if len(trace_class.data_frame) < 1:
             return
-
+        # Get events DF boundaries
+        min_t = trace_class.data_frame.index[0]
+        max_t = trace_class.data_frame.index[-1]
+        # Restrict events DF boundaries according to the specified window
+        if window[0]:
+            max_t = min(max_t, window[1])
```
valschneider: AFAICT `window[1]` can be `None` and

```python
>>> print min(42, 43)
42
>>> print min(42, None)
None
```

valschneider:
@derkling I forgot about this PR. I think https://github.com/ARM-software/trappy/pull/279 fixes the error you were seeing.

PR #278, opened 2018-04-25
Travis fixes

valschneider:


.travis.yml
```diff
@@ -19,11 +19,5 @@ env:
 script: nosetests
 virtualenv:
   system_site_packages: true
-notifications:
```
JaviMerino: No, thanks.  I'd rather keep receiving these emails.

tests/test_copyright.py
```diff
@@ -48,11 +50,18 @@ def copyright_is_valid(fname):
     if "Google Inc" in lines[1]:
         apache_line += 1
 
-    # The Copyright includes the current year
+    # The Copyright includes valid years
     current_year = date.today().year
-    if str(current_year) not in lines[0]:
+    re_years = re.compile(r"[-\s](?P<year>\d+)")
+    years = map(int, re_years.findall(lines[0]))
```
JaviMerino: You don't get anything by compiling the regexp.  It's more efficient and compact to just run `re.findall` on the line:

    years = map(int, re.findall(r"[-\s](\d+)", lines[0]))

tests/test_copyright.py
```diff
@@ -20,6 +20,8 @@
 import re
 import unittest
 
+RE_YEAR = re.compile(r"")
```
JaviMerino: This is not used and not needed.

valschneider: Woopsie daisy, forgot about that one there...

PR #282, opened 2018-09-05
Remove print event

douglas-raillard-arm:
Remove Print event class since it interferes with "tracing_mark_write" matching. Print event has been introduced very recently in the library and no external code base is using it yet.

Also add a sanity check on unique words to make sure that they are never a substring of another one.

### DIFF
--- a/tests/test_caching.py
+++ b/tests/test_caching.py
@@ -248,7 +248,7 @@ class TestCaching(utils_tests.SetupDirectory):
         trace_dir = os.path.dirname(trace_path)
         trace_file = os.path.basename(trace_path)
         cache_dir = '.' + trace_file + '.cache'
-        number_of_trace_categories = 32
+        number_of_trace_categories = 31
         self.assertEquals(len(os.listdir(cache_dir)), number_of_trace_categories)
 
         os.remove(os.path.join(cache_dir, 'SchedWakeup.csv'))
--- a/tests/test_ftrace.py
+++ b/tests/test_ftrace.py
@@ -312,7 +312,7 @@ class TestFTrace(BaseTestThermal):
     def test_plot_allfreqs_with_one_actor(self):
         """Check that plot_allfreqs() works with one actor"""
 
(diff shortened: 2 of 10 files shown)


tests/test_ftrace.py
```diff
@@ -385,7 +385,7 @@ def test_parse_tracing_mark_write_events(self):
         # put under trace.cpu_frequency
         self.assertEquals(trace.tracing_mark_write.data_frame.iloc[-1]["string"],
                           "TRACE_MARKER_START")
-        self.assertEquals(len(trace.tracing_mark_write.data_frame), 1)
```
JaviMerino: No, the code and the comment are already aligned.  The comment says that there should be only one event in `tracing_mark_write` because the second one should be in `cpu_frequency` and this test checks that it is true

tests/test_caching.py
```diff
@@ -248,7 +248,7 @@ def test_cache_delete_single(self):
         trace_dir = os.path.dirname(trace_path)
         trace_file = os.path.basename(trace_path)
         cache_dir = '.' + trace_file + '.cache'
-        number_of_trace_categories = 32
+        number_of_trace_categories = 31
```
JaviMerino: this is weird.  You are not changing the inputs to this test, so why did you have to reduce the number of trace categories?

tests/test_ftrace.py
douglas-raillard-arm: Ok, I'll add the check on both then, since we also want to make sure that cpu_frequency got its item.

tests/test_caching.py
douglas-raillard-arm: From what I could see, there was one CSV file per event class name. Since I deleted the "Print" class, this one disappeared.

JaviMerino: ah, ok.  What confused me is that it was in the commit "event: rename thermal_power_cpu_get" instead of "events: remove print event".  Makes sense now.

PR #284, opened 2018-09-20
ftrace: fix __getattr__ implementation

douglas-raillard-arm:
Since `__getattr__` can be called on uninitialized instances (i.e. with no
attributes), it can enter an infinite recursion while trying to access
some instance attributes.

Fix that by ignoring requests for dunder names (`__XXX__`), since they are
resolved without `__getattr__` when needed by a builtin, and are mostly
supposed to be class attributes in other cases. That means `__getattr__`
is not supposed to receive requests to such names often, and these will
(almost) surely not be anything the API can offer since usage of such
names is reserved for future expansion of the language.

### DIFF
--- a/trappy/ftrace.py
+++ b/trappy/ftrace.py
@@ -488,27 +488,31 @@ is part of the trace.
     def __getattr__(self, attr):
         """Raises useful exception when trying to access deprecated
         attributes."""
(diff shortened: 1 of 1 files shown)


JaviMerino:
Tests pass on my machin, so I ignoring travis' test failure

PR #286, opened 2018-09-28
trappy: utils: Fix get_duplicates FutureWarning

valschneider:
We've had this one for a few months now:

trappy/utils.py:78: FutureWarning: 'get_duplicates' is deprecated and will be removed in a future release. You can use idx[idx.duplicated()].unique() instead
  dups = index.get_duplicates()

Do what the warning says.

### DIFF
--- a/trappy/utils.py
+++ b/trappy/utils.py
@@ -75,7 +75,7 @@ def handle_duplicate_index(data,
     index = data.index
     new_index = index.values
 
-    dups = index.get_duplicates()
+    dups = index[index.duplicated()].unique()
 
     for dup in dups:
         # Leave one of the values intact
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ based on the underlying FTrace data.
 REQUIRES = [
     "numpy",
     "pyparsing",
-    "pandas>=0.13.1",
+    "pandas>=0.15.0",
     "future",
 ]
 


valschneider:
(Woopsie daisy, this is straight for my lisa tree, forgot to remove the trappy: prefix, hang on)

trappy/utils.py
```diff
@@ -75,7 +75,7 @@ def handle_duplicate_index(data,
     index = data.index
     new_index = index.values
 
-    dups = index.get_duplicates()
+    dups = index[index.duplicated()].unique()
```
JaviMerino: `unique()` was introduced in pandas 0.23.0.  Can you bump the pandas version requirement in `setup.py` to 0.23.0 as part of this commit please?

valschneider: Will do!

valschneider: Mmm wouldn't 0.15.0 be enough? I have nothing against 0.23.0, but looking at the API doc it seems that:
* `Index.unique()` was introduced in 0.13.0 (https://pandas.pydata.org/pandas-docs/version/0.13/generated/pandas.Index.unique.html)
* `Index.duplicated()` was introduced in 0.15.0 (https://pandas.pydata.org/pandas-docs/version/0.15/generated/pandas.Index.duplicated.html)

JaviMerino: Ah, right, I misread the documentation.  The `level` parameter is the one that was added in 0.23.0.  You are correct, 0.15.0 is enough.

valschneider: Do we want to take the opportunity to bump up to a higher version though? 0.15.0 is from 2014, which is getting old.

JaviMerino: We are setting the lower bound (`pandas>=0.15.0`), no need to bump it to something newer without a reason.

valschneider: Okay, works for me

PR #287, opened 2018-11-01
trappy/cache: Store the csv in utf-8 charset explicitly

credp:
When we load the files from trace-cmd output, we explicitly choose
to load them using utf-8. This means we may have characters in
our trace which cannot be represented in ascii, which makes pandas
upset when we try to store the parsed dataframes in the csv cache.

This is not an issue for Python3 versions, as the to_csv method
defaults to utf-8, whilst the Python2 to_csv method defaults to
ascii.

Signed-off-by: Chris Redpath <chris.redpath@arm.com>

### DIFF
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -298,7 +298,7 @@ class Base(object):
         :param fname: The name of the CSV file
         :type fname: str
         """
-        self.data_frame.to_csv(fname)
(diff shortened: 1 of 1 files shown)


valschneider:
Travis doesn't seem to like, can you have a look as to why?

PR #288, opened 2018-12-19
base: Fix string to int conversion

valschneider:
Because it relied on str.isdigit(), Base.string_cast() would
erroneously assume a hex string (e.g. "0x42") wouldn't be castable to
int.

Use a more robust pattern using try/except blocks. This will happily
cope with negative and hex values.

While at it, rename string_cast() to string_cast_int(), since the
former only handles ints and does an ugly empty return when the
specified type isn't int.

### DIFF
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -190,26 +190,21 @@ class Base(object):
         self.line_array.append(line)
         self.data_array.append(data)
 
-    def string_cast(self, string, type):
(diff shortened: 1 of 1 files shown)


valschneider:
We seem to have been hitting this "KeyError: 'A53_freq_out'" travis error for a while. It seems to be with Python3 only, any ideas why?

PR #290, opened 2019-04-03
trace/base: Optimizing DataFrame memory footprint

derkling:
Under the hood pandas represents numeric values as NumPy ndarrays and
stores them in a continuous block of memory. Values of the same column
are represented using the same type and thus number of bytes.

Many types in pandas have multiple subtypes that can use fewer bytes to
represent each value. For example, the float type has the float16,
float32, and float64 subtypes.

Use the function pd.to_numeric() to downcast numeric types to use for
each value the minumum number of bytes which is still enough to
represent the maximum value for a given column.

Use also "Categoricals" introduced in Pandas since version 0.15.
The category type uses integer values under the hood to represent the
values in a column, rather than the raw values. Use category to
efficiently compress the representation of string values by replacing
64bit string pointers with an index using less bits.

Credits goes to:

   Using pandas with Large Data Sets
   https://www.dataquest.io/blog/pandas-big-data/

where these changes are proposed and discussed in details.

The proposed change applied to a 473M example trace gives the following
results:

                   | Events |  Memory (MB)  |  Compression
                   |  count | Before  After |      percent
  -----------------+--------+ --------------+-------------
   clock_disable   |  18368 |   3.34   0.88 |    73.652695
   clock_enable    |  19149 |   3.43   0.92 |    73.177843
   clock_set_rate  |  42099 |   7.63   2.01 |    73.656619
   cpu_idle        | 272726 |  28.87  12.74 |    55.871147
   sched_switch    | 315951 |  82.86  23.26 |    71.928554

### DIFF
--- a/trappy/base.py
+++ b/trappy/base.py
@@ -265,6 +265,27 @@ class Base(object):
 
             yield data_dict
 
+    def optimize_dataframe(self):
+        """Optimize memory footprint by setting minimal data types required by
+           each column"""
+        for col in self.data_frame.columns:
+            if self.data_frame[col].dtype.kind == 'i':
+                self.data_frame[col].apply(pd.to_numeric, downcast='signed')
+                continue
+            if self.data_frame[col].dtype.kind == 'f':
+                self.data_frame[col].apply(pd.to_numeric, downcast='float')
+                continue
+            if self.data_frame[col].dtype.kind == 'S':
+                # Convert string objects (pointer) to categories, only when we have
+                # a relatively limited number of unique values (50% of the rows)
+                num_unique_values = len(self.data_frame[col].unique())
+                num_total_values = len(self.data_frame[col])
+                if num_unique_values / num_total_values > 0.5:
+                    continue
+                self.data_frame.loc[:,col] = self.data_frame[col].astype('category')
+            else:
+                continue
+
     def create_dataframe(self):
         """Create the final :mod:`pandas.DataFrame`"""
         if not self.time_array:
@@ -280,6 +301,7 @@ class Base(object):
         time_idx = pd.Index(self.time_array, name="Time")
         self.data_frame = pd.DataFrame(self.generate_parsed_data(), index=time_idx)
         self.data_frame = handle_duplicate_index(self.data_frame)
+        self.optimize_dataframe()
 
         self.time_array = []
         self.line_array = []
@@ -309,6 +331,7 @@ class Base(object):
             # same method, aka python's float() and not numpy's
             converters={'Time' : float}
         )
+        self.optimize_dataframe()
 
     def normalize_time(self, basetime):
         """Substract basetime from the Time of the data frame


trappy/base.py
```diff
@@ -265,6 +265,24 @@ def generate_parsed_data(self):
...
+                self.data_frame[col].apply(pd.to_numeric, downcast='unsigned')
+                continue
+            if 'float' in str(self.data_frame[col].dtype):
+                self.data_frame[col].apply(pd.to_numeric, downcast='float')
+                continue
+            # Convert string objects (pointer) to categories, only when we have
+            # a relatively limited number of unique values (50% of the rows)
+            num_unique_values = len(self.data_frame[col].unique())
+            num_total_values = len(self.data_frame[col])
+            if num_unique_values / num_total_values > 0.5:
```
valschneider: Mmm you can argue that you can save some memory even when you have >50% unique names, but that's not too relevant for us, I think.

trappy/base.py
```diff
@@ -265,6 +265,24 @@ def generate_parsed_data(self):
 
             yield data_dict
 
+    def optimize_dataframe(self):
+        """Optimize memory footprint by setting minimal data types required by
+           each column"""
+        for col in self.data_frame.columns:
+            if 'int' in str(self.data_frame[col].dtype):
+                self.data_frame[col].apply(pd.to_numeric, downcast='unsigned')
```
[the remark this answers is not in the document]
derkling: Mmm... right... actually, I'm wondering if we cannot just skip the check and always downcast to signed. Should not imply using more bits in most cases.
Do you think we rely on unsigned in some cases?

trappy/base.py
derkling: Right, that's a very coarse-grained heuristic... in the post linked by the commit message the guy suggest this value as something conservative. Still, I think it quite well fit most of our use cases where, with trace events, we have tons of repeated strings (e.g. task names).

derkling:
Regarding the need to firsts build the DF I agree. However, this patch main goal is to reduce the memory footprint for live traces and (perhaps) speed-up a bit the analysis we run once the DFs are in memory... although I did not measure that.

A relatively simple extension could be to store the columns dtypes in the cache, that would allow loading the CSV with the correct dtype starting from the second usage of a trace.

trappy/base.py
```diff
@@ -265,6 +265,24 @@ def generate_parsed_data(self):
...
+                continue
+            if 'float' in str(self.data_frame[col].dtype):
```
douglas-raillard-arm: Checking the dtype kind seems more appropriate, since dtypes may be more complex and likely include "float" in their human-readable string representation, although they are not floats at all (same goes for int).
```
self.data_frame[col].kind == 'f'
```

trappy/base.py
```diff
@@ -265,6 +265,24 @@ def generate_parsed_data(self):
...
+        """Optimize memory footprint by setting minimal data types required by
+           each column"""
+        for col in self.data_frame.columns:
+            if self.data_frame[col].dtype.kind == 'i':
+                self.data_frame[col].apply(pd.to_numeric, downcast='signed')
+                continue
+            if self.data_frame[col].dtype.kind == 'f':
+                self.data_frame[col].apply(pd.to_numeric, downcast='float')
+                continue
+            # Convert string objects (pointer) to categories, only when we have
```
valschneider: Shouldn't we do this?
```
if self.data_frame[col].dtype.kind == 'S':
    # string optimization here
else:
    continue

Comment (derkling) [reply trappy/base.py]: Well, for trace events we have either integers, floats or strings... do we have other values to consider?

Comment (derkling) [reply trappy/base.py]: There are arrays... but I'm not sure how they are represented and parsed by trappy.
@JaviMerino should know it better...

Comment (valschneider) [reply trappy/base.py]: I seem to remember some thermal event parsing being somewhat botched turning cpumask into a weird type (or maybe it might just have turned them into weird looking strings).

Comment (JaviMerino) [reply trappy/base.py]: arrays in a trace are expanded. `load=[0, 20, 42, 50]` becomes `load0=0 load1=20 load2=42 load3=50`. So the dataframe will see them as columns with numbers and you optimizations apply.

File: trappy/base.py [hunk 5]
```diff
@@ -265,6 +265,24 @@ def generate_parsed_data(self):
...
+            # a relatively limited number of unique values (50% of the rows)
+            num_unique_values = len(self.data_frame[col].unique())
```
JaviMerino: Interesting. I would have done `len(set(self.data_fram[col]))`, but I guess if the dataframe has a `.unique()` method it has to be faster than `set()`.

trappy/base.py
JaviMerino: I'd be surprised if in sufficiently large dataframes you have this over the 50% mark to be honest

derkling:
Is anyone brave enough to merge this one? :)

valschneider:
Huh so apparently we get this DataFrame:
```
Time
0.000102     NaN
0.000110     NaN
0.674103    None
0.674115    None
0.678762    None
0.678785    None
0.679127    None
0.681465    None
0.681766       1
0.681778    None
0.681787    None
0.683561       1
0.683991       1
0.685980    None
0.686025    None
0.686068    None
0.686093    None
0.686186    None
0.687742    None
0.687784    None
0.688318    None
0.688932    None
0.688983    None
0.689777    None
0.690072    None
0.690085    None
0.690120    None
0.691775    None
0.691787    None
0.691881    None
            ... 
Name: data, Length: 160, dtype: object
```

And `self.assertTrue(isnan(dfr['data'].iloc[2]))` is failing because we get a `None`, not a `NaN`... Seems like the test code before your change could handle that. So this might have been a case where we used to do the string optimization for !string columns (dtype is just `object`).

valschneider:
Right, and with the previous version (no `else: continue`) we get this dataframe:
```
Time
0.000102    NaN
0.000110    NaN
0.674103    NaN
0.674115    NaN
0.678762    NaN
0.678785    NaN
0.679127    NaN
0.681465    NaN
0.681766      1
0.681778    NaN
0.681787    NaN
0.683561      1
0.683991      1
0.685980    NaN
0.686025    NaN
0.686068    NaN
0.686093    NaN
0.686186    NaN
0.687742    NaN
0.687784    NaN
0.688318    NaN
0.688932    NaN
0.688983    NaN
0.689777    NaN
0.690072    NaN
0.690085    NaN
0.690120    NaN
0.691775    NaN
0.691787    NaN
0.691881    NaN
           ... 
Name: data, Length: 160, dtype: category
Categories (5, object): [0, 1, 2, 25601320, 37]
```

So I *think* the right thing to do is to keep that `else: continue` but revert the changes to the tests.

derkling:
@qais-yousef  changes went in before... lets to the DF optimization once the index has been fixed ;)

qais-yousef:
I've created this pull request to pull the new changes in trappy to lisa

https://github.com/ARM-software/lisa/pull/890

PR #291, opened 2019-04-12
Hex conversion downtuning

valschneider:
Fixes a bug reported by @derkling, details in the commit log.


valschneider:
I realized that this initial version didn't accept things like "-0x42". Then I looked at the doc of `int()` and found this:
```
Base 0 means to interpret exactly as a code literal,
```

So I'm just doing `int(string, base=0)` now.

derkling:
LGTM

PR #297, opened 2019-08-19
Fixed error when parsing android user mark

Pablololo12:
I found out that when you use the Android tools to create trace markers from within the app, it doesn't recognize the PID of the app on the Trace.endSection() due to the regex. With this fix you can see when a trace marker ends.

### DIFF
--- a/trappy/systrace.py
+++ b/trappy/systrace.py
@@ -21,7 +21,7 @@ from trappy.ftrace import GenericFTrace
 import re
 
 SYSTRACE_EVENT = re.compile(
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
LGTM, waiting for an extra ack from @JaviMerino or @derkling

JaviMerino:
I'm missing something. The unittest for `tracing_mark_write` in tests/test_systrace.py:79 doesn't expect the PID because the trace's end events don't have pids.  Looking at the test trace, I see

`surfaceflinger-594 ( 594) [002] ...1   538.785992: tracing_mark_write: E`

Are you seeing something different? If so, can you add it to the test?

PR #300, opened 2019-11-14
ftrace: Only add dynamic classes on non-custom scopes

douglas-raillard-arm:
Custom scope is supposed to give only the events that are explicitly listed in
`events`. Therefore, avoid adding dynamic events in this case, to speed up
parsing and avoid consuming memory.


douglas-raillard-arm:
> Could you fix the tests and doc/Dynamic traces.ipynb

Do you mean converting Python 3 code to Python 2 ? Other than that the way the Ftrace is used should stay the same, as scope defaults to `"all"`, which is `!= 'custom'` and will therefore get the dynamic parsers.

tests/test_ftrace.py
```diff
@@ -482,9 +482,7 @@ def test_ftrace_unique_but_no_fields(self):
         with open("trace.txt", "a") as fil:
             fil.write("version = 6")
 
-        with self.assertRaises(ValueError):
-            trappy.FTrace(scope="custom")
-
+        trappy.FTrace(scope="custom")
```
JaviMerino: No, you have changed the meaning of custom to avoid having the dynamically registered classes. The proper fix for this test is:
``` python 
with self.assertRaises(ValueError):
            trappy.FTrace(scope="all")
```

douglas-raillard-arm:
@JaviMerino Updated with unit test fix

